PR #55, opened 2015-02-26
Better 2/3 compatible string type checking

thorstenkranz:
FIX: unicode strings won't be recognized as string formulas in Python 2.x

### DIFF
--- a/patsy/highlevel.py
+++ b/patsy/highlevel.py
@@ -13,6 +13,7 @@ __all__ = ["dmatrix", "dmatrices",
 #   ModelDesign doesn't work -- need to work with the builder set
 #   want to be able to return either a matrix or a pandas dataframe
 
+from six import string_types
 import numpy as np
 from patsy import PatsyError
 from patsy.design_info import DesignMatrix, DesignInfo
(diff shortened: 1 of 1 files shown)


njsmith:
Unfortunately, this change is incorrect, for the reasons described here:
http://patsy.readthedocs.org/en/latest/py2-versus-py3.html
I'd like to support Unicode strings in py2, but it's just not possible to
do correctly, because patsy formulas contain Python code (all the variable
names etc.), and in py2, *Python code itself* cannot be written in Unicode,
and there's nothing Patsy can do to change this. (Except switch to py3 of
course ;-).)

The confusing error message you got is definitely a bug though -- a patch
to give better diagnostics in this case would be welcome.

thorstenkranz:
Thank you for your really quick reply. I understand, and am ashamed I
didn't see this in the documentation.

Greetings,
Thorsten

2015-02-26 19:47 GMT+01:00 Nathaniel J. Smith <notifications@github.com>:

thorstenkranz:
I issued another pull request, now with a more verbose error message. HTH.

2015-02-26 20:25 GMT+01:00 Thorsten Kranz <thorstenkranz@googlemail.com>:

njsmith:
No need to be ashamed! It is a pretty obscure and annoying issue :-/

PR #56, opened 2015-02-26
Create more verbose error message for unicode strings as formulas

thorstenkranz:
Instead of supporting unicode, give a useful error message when a formula is unicode.

### DIFF
--- a/patsy/highlevel.py
+++ b/patsy/highlevel.py
@@ -13,6 +13,7 @@ __all__ = ["dmatrix", "dmatrices",
 #   ModelDesign doesn't work -- need to work with the builder set
 #   want to be able to return either a matrix or a pandas dataframe
 
+import six
 import numpy as np
 from patsy import PatsyError
 from patsy.design_info import DesignMatrix, DesignInfo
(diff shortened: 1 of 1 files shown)


patsy/highlevel.py
```diff
@@ -56,6 +57,9 @@ def _try_incr_builders(formula_like, data_iter_maker, eval_env,
                                       data_iter_maker,
                                       NA_action)
     else:
+        # check if formula_like is another string type (which might be unsupported)
+        if isinstance(formula_like, string_types):
+            raise PatsyError("Unsupported string type for formula (e.g., unicode in Python 2.X).")
```
njsmith: I think this is written in a little confusing way -- the only posible cases are:
* Python 2: `string_types` is equivalent to `(str, unicode)`. `str` is handled above, so only `unicode` can hit this point.
* Python 3: `string_types` is equivalent to `str`. `str` is handled above, so this point can't be reached.
There aren't any other string types floating around.

So I'd probably write the check like `if not six.PY3 and isinstance(formula_like, unicode): ...`, and then the error message could be more specific. (Unfortunately you can't use `if six.PY2 ...` because apparently older versions of `six` define `PY3` but not `PY2`... I got another bug report about this at some point.)

njsmith:
Also, could you add a test for this? We try to maintain ~100% line coverage. Just something simple like `assert_raises_regexp(PatsyError, "unicode", dmatrix, u"some formula that used to produce a confusing error", {})`

njsmith:
Closing in favor of #76.

PR #65, opened 2015-04-15
Also flag test_crs_compat as a slow test.

chrish42:


njsmith:
New milestone: I'm pretty sure this is the first PR from someone who isn't me, which modifies code written by someone else who isn't me :-)

Issue #25, opened 2013-10-06
Capture only the values of referenced variables in formula namespace


njsmith:
Right now when creating a formula, we capture the namespace itself.

This can pin large variables in memory, and presents an obstacle to serializiing model designs (#26).

What we should do is to figure out which variables from the enclosing namespace are actually used, and then capture only those.

The klugey way to do this is to observe which variables are accessed when evaluating the formula the first time, and then save only those.

The more principle, reliable, and modular way is to use `ast` to parse the formula, and then extract all bare variable references. (Or maybe we should just re-use the token-based implementation of this.) Those which are found not in the data, and not in the builtins, but in the environment, should get stashed to use for actual evaluation.

This isn't just an optimization, it does produce a user-visible effect: if some variable name referenced in the formula is rebound after the formula is created, then previously the new value would be used in future predictions, but after this change, the old value will continue to be used.

This is unaesthetic (I think PHP's so-called "closures" work this way?), but in our case it's actually for the best -- ideally we'd save a read-only snapshot of the environment, period, which is not tractable. But this moves us slightly in that direction, so, okay.

njsmith:
Based on discussion with @chrish42 at PyCon, I think the basic idea is: Inside `design_matrix_builders`, which "sniffing" the data, we should check which global variables are referenced inside each factor (e.g. `"x"` references `x`, and `"np.log(y)"` references `np` and `y`), and we should make a note for each variable whether it is found in (a) the 'training data', (b) the captured environment, or (c) the builtins. And in case (b) (and possibly also c) we should write down what the value actually was. (This would all go into the factor `state`.) Then later, at the `build_design_matrices` stage, we should only do lookups using this `state`, not the actual environment.

(One side-effect is that this would mean that some corner cases would be handled in a nicer way. E.g. if you have both a global variable called "x" and a variable in your data frame called "x", then right now the initial building will look at the x in the data frame -- but if later when predicting you accidentally forget to give an x variable, you will silently get the global variable x instead. With this change, it will instead error out, insisting that anything that was in the data frame can only come from the data frame.)

[...]

Issue #13, opened 2013-01-09
Trouble making categorical variable --- TypeError: 'ClassRegistry' object is not callable


selik:
Not sure what's going on here.

```
In [44]: patsy.dmatrix("C(a)", {'a':['m', 'n', 'o']})
Out[44]:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-44-24c8bf15ba3b> in <module>()
----> 1 dmatrix("C(a)", {'a':['m', 'n', 'o']})

/Users/mike/venv/sci/lib/python2.7/site-packages/patsy/highlevel.pyc in dmatrix(formula_like, data, eval_env, return_type)
    259     """
    260     (lhs, rhs) = _do_highlevel_design(formula_like, data, _get_env(eval_env),
--> 261                                       return_type)
    262     if lhs.shape[1] != 0:
    263         raise PatsyError("encountered outcome variables for a model "

/Users/mike/venv/sci/lib/python2.7/site-packages/patsy/highlevel.pyc in _do_highlevel_design(formula_like, data, eval_env, return_type)
    145     def data_iter_maker():
    146         return iter([data])
--> 147     builders = _try_incr_builders(formula_like, data_iter_maker, eval_env)
    148     if builders is not None:
    149         return build_design_matrices(builders, data,

/Users/mike/venv/sci/lib/python2.7/site-packages/patsy/highlevel.pyc in _try_incr_builders(formula_like, data_iter_maker, eval_env)
     59         return design_matrix_builders([formula_like.lhs_termlist,
     60                                        formula_like.rhs_termlist],
---> 61                                       data_iter_maker)
     62     else:
     63         return None

/Users/mike/venv/sci/lib/python2.7/site-packages/patsy/build.pyc in design_matrix_builders(termlists, data_iter_maker)
    691      cat_postprocessors) = _examine_factor_types(all_factors,
    692                                                  factor_states,
--> 693                                                  data_iter_maker)
    694     # Now we need the factor evaluators, which encapsulate the knowledge of
    695     # how to turn any given factor into a chunk of data:

/Users/mike/venv/sci/lib/python2.7/site-packages/patsy/build.pyc in _examine_factor_types(factors, factor_states, data_iter_maker)
    441             break
    442         for factor in list(examine_needed):
--> 443             value = factor.eval(factor_states[factor], data)
    444             if isinstance(value, Categorical):
    445                 postprocessor = CategoricalTransform(levels=value.levels)

[...]

PR #66, opened 2015-04-16
Capture vars not env

chrish42:
This is the current state of my work  to capture only the required variables for patsy formulas instead of the whole environment.

Current status: 
- [x] Add new methods to EvalEnvironment to create a environment that contains only a subset of variables.
- [x] Refactor EvalFactor (and update all users) to move the eval_env parameter out of the its initializer method and into the state.
- [x] Additions to doc/changes.rst (esp. to document the incompatible changes!)
- [x] Do a quick pass through the docs to catch anything that's now out-of-date.
- [x] Add "end-to-end" test or two -- something that tests the user-visible behaviour directly, like running a formula through dmatrix, then modifying the original environment, and then running some more data through the same builder (like we were doing predictions) and checking that it is unaffected by our changes to the env. (test_highlevel.py is where the other tests like this go.)
- [x] Create right subset EvalEnvironment and stash it into the state dict for EvalFactor.
- [x] ~~Figure out what the right thing to do is when variables are shadowed between the EvalFactor's environment and the data.~~ (Can be done together with #13 instead. Not a blocker for this.)

When this is complete, it will fix bug #25.

### DIFF
--- a/patsy/build.py
+++ b/patsy/build.py
@@ -324,14 +324,14 @@ def test__ColumnBuilder():
     cb_intercept.build({f1: [1, 2, 3], f2: [1, 2, 3], f3: [1, 2, 3]}, mat3)
     assert np.allclose(mat3, 1)
 
-def _factors_memorize(factors, data_iter_maker):
+def _factors_memorize(factors, data_iter_maker, eval_env):
     # First, start off the memorization process by setting up each factor's
     # state and finding out how many passes it will need:
     factor_states = {}
     passes_needed = {}
     for factor in factors:
         state = {}
-        which_pass = factor.memorize_passes_needed(state)
+        which_pass = factor.memorize_passes_needed(state, eval_env)
         factor_states[factor] = state
         passes_needed[factor] = which_pass
     # Now, cycle through the data until all the factors have finished
@@ -361,7 +361,7 @@ def test__factors_memorize():
             self._chunk_in_pass = 0
             self._seen_passes = 0
 
-        def memorize_passes_needed(self, state):
+        def memorize_passes_needed(self, state, eval_env):
             state["calls"] = []
             state["token"] = self._token
             return self._requested_passes
@@ -388,7 +388,7 @@ def test__factors_memorize():
     f1 = MockFactor(1, "f1")
     f2a = MockFactor(2, "f2a")
     f2b = MockFactor(2, "f2b")
-    factor_states = _factors_memorize(set([f0, f1, f2a, f2b]), data)
+    factor_states = _factors_memorize(set([f0, f1, f2a, f2b]), data, {})
     assert data.calls == 2
     mem_chunks0 = [("memorize_chunk", 0)] * data.CHUNKS
     mem_chunks1 = [("memorize_chunk", 1)] * data.CHUNKS
@@ -614,7 +614,7 @@ def _make_term_column_builders(terms,
             term_to_column_builders[term] = column_builders
     return new_term_order, term_to_column_builders
 
-def design_matrix_builders(termlists, data_iter_maker, NA_action="drop"):
+def design_matrix_builders(termlists, data_iter_maker, eval_env=0, NA_action="drop"):
     """Construct several :class:`DesignMatrixBuilders` from termlists.
 
     This is one of Patsy's fundamental functions. This function and
@@ -628,6 +628,14 @@ def design_matrix_builders(termlists, data_iter_maker, NA_action="drop"):
       simple iterator because sufficiently complex formulas may require
       multiple passes over the data (e.g. if there are nested stateful
       transforms).
+    :arg eval_env: Either a :class:`EvalEnvironment` which will be used to
+      look up any variables referenced in `termlists` that cannot be
+      found in `data_iter_maker`, or else a depth represented as an
+      integer which will be passed to :meth:`EvalEnvironment.capture`.
+      ``eval_env=0`` means to use the context of the function calling
+      :func:`design_matrix_builders` for lookups. If calling this function
+      from a library, you probably want ``eval_env=1``, which means that
+      variables should be resolved in *your* caller's namespace.
     :arg NA_action: An :class:`NAAction` object or string, used to determine
       what values count as 'missing' for purposes of determining the levels of
       categorical factors.
@@ -642,14 +650,20 @@ def design_matrix_builders(termlists, data_iter_maker, NA_action="drop"):
 
     .. versionadded:: 0.2.0
        The ``NA_action`` argument.
+    .. versionadded:: 0.4.0
+       The ``eval_env`` argument.
     """
+    if not isinstance(eval_env, six.integer_types + (EvalEnvironment,)):
+        raise TypeError("Parameter 'eval_env' must be either an integer or an instance" +
+                        "of patsy.EvalEnvironment.")
+    eval_env = EvalEnvironment.capture(eval_env, reference=1)
     if isinstance(NA_action, str):
         NA_action = NAAction(NA_action)
     all_factors = set()
     for termlist in termlists:
         for term in termlist:
             all_factors.update(term.factors)
-    factor_states = _factors_memorize(all_factors, data_iter_maker)
+    factor_states = _factors_memorize(all_factors, data_iter_maker, eval_env)
     # Now all the factors have working eval methods, so we can evaluate them
     # on some data to find out what type of data they return.
     (num_column_counts,
--- a/patsy/build.py
+++ b/patsy/build.py
@@ -614,7 +614,7 @@ def _make_term_column_builders(terms,
             term_to_column_builders[term] = column_builders
     return new_term_order, term_to_column_builders
 
-def design_matrix_builders(termlists, data_iter_maker, eval_env=0, NA_action="drop"):
+def design_matrix_builders(termlists, data_iter_maker, eval_env, NA_action="drop"):
     """Construct several :class:`DesignMatrixBuilders` from termlists.
 
     This is one of Patsy's fundamental functions. This function and
--- a/patsy/build.py
+++ b/patsy/build.py
@@ -659,7 +659,7 @@ def design_matrix_builders(termlists, data_iter_maker, eval_env, NA_action="drop
     # not as a keyword argument a nice error message here, instead of a
     # more obscure backtrace later on.
     if not isinstance(eval_env, six.integer_types + (EvalEnvironment,)):
-        raise TypeError("Parameter 'eval_env' must be either an integer or an instance" +
+        raise TypeError("Parameter 'eval_env' must be either an integer or an instance "
                         "of patsy.EvalEnvironment.")
     eval_env = EvalEnvironment.capture(eval_env, reference=1)
     if isinstance(NA_action, str):
--- a/patsy/build.py
+++ b/patsy/build.py
@@ -653,6 +653,11 @@ def design_matrix_builders(termlists, data_iter_maker, eval_env=0, NA_action="dr
     .. versionadded:: 0.4.0
        The ``eval_env`` argument.
     """
+    # Check type of eval_env to help people migrating to 0.4.0. Third
+    # argument used to be NA_action (a string). Having the check for
+    # eval_env's type gives people migrating to 0.4.0 who used NA_action
+    # not as a keyword argument a nice error message here, instead of a
+    # more obscure backtrace later on.
     if not isinstance(eval_env, six.integer_types + (EvalEnvironment,)):
         raise TypeError("Parameter 'eval_env' must be either an integer or an instance" +
                         "of patsy.EvalEnvironment.")
--- a/patsy/desc.py
+++ b/patsy/desc.py
@@ -584,7 +584,7 @@ _eval_error_tests = [
     "a + <-a**2>",
 ]
 
-def _assert_terms_match(terms, expected_intercept, expecteds, eval_env): # pragma: no cover
+def _assert_terms_match(terms, expected_intercept, expecteds): # pragma: no cover
     if expected_intercept:
         expecteds = [()] + expecteds
     assert len(terms) == len(expecteds)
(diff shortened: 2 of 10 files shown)


patsy/build.py
```diff
@@ -642,14 +650,25 @@ def design_matrix_builders(termlists, data_iter_maker, NA_action="drop"):
...
+       The ``eval_env`` argument.
     """
+    # Check type of eval_env to help people migrating to 0.4.0. Third
+    # argument used to be NA_action (a string). Having the check for
+    # eval_env's type gives people migrating to 0.4.0 who used NA_action
+    # not as a keyword argument a nice error message here, instead of a
+    # more obscure backtrace later on.
+    if not isinstance(eval_env, six.integer_types + (EvalEnvironment,)):
+        raise TypeError("Parameter 'eval_env' must be either an integer or an instance" +
+                        "of patsy.EvalEnvironment.")
```
njsmith: Nitpicks:
* The "+" is unnecessary here -- the Python lexer (like that in many languages, e.g. C) will concatenate adjacent string tokens at compile time, e.g. `"foo" "bar"` is the same as `"foobar"` (and so is `"foo" 'bar'` for that matter). It's even slightly more efficient, b/c `+` is evaluated at runtime, while leaving it out gives a single string directly in the .pyc. Not that this efficiency matters at all :-). But as a style/consistency thing I leave the `+` out.
* You're missing a space, this string says "instanceof" :-)

patsy/desc.py
```diff
@@ -593,8 +592,7 @@ def _assert_terms_match(terms, expected_intercept, expecteds, eval_env): # pragm
         if isinstance(term, Term):
             if isinstance(expected, str):
                 expected = (expected,)
-            assert term.factors == tuple([EvalFactor(s, eval_env)
-                                          for s in expected])
+            assert term.factors == tuple([EvalFactor(s) for s in expected])
```
njsmith: I think `eval_env` can probably be removed more thoroughly from this code? (I.e., if there aren't any other users in the function, then the variable can be removed entirely, and so on up the call stack.)

patsy/eval.py
```diff
@@ -81,6 +82,42 @@ def test_VarLookupDict():
...
     assert d1["a"] == 1
+    assert ds.get("c") is None
+    assert isinstance(repr(ds), six.string_types)
+
+def ast_names(code):
+    """Iterator that yields all the (ast) names in a Python expression.
+
+    :arg code: A string containing a Python expression.
+    """
+    disallowed_ast_nodes = (ast.Lambda, ast.ListComp, ast.GeneratorExp)
```
njsmith: Let's add a comment here saying something like "# Syntax that allows new name bindings to be introduced is tricky to handle here, so we just refuse to do so"

patsy/eval.py
```diff
@@ -81,6 +82,42 @@ def test_VarLookupDict():
...
+    if sys.version_info >= (2, 7):
+        disallowed_ast_nodes += (ast.DictComp, ast.SetComp)
+
+    for node in ast.walk(ast.parse(code)):
+        if isinstance(node, disallowed_ast_nodes):
+            raise PatsyError("Lambda, list/dict/set comprehension, generator expression in patsy formula not currently supported.")
```
njsmith: Please break this line up -- we try to stick below 80 characters per line.

patsy/eval.py
```diff
@@ -460,16 +522,16 @@ def new_name_maker(token):
...
         return call_and_wrap_exc("Error evaluating factor",
                                  self,
-                                 self._eval_env.eval,
+                                 eval_env.eval,
                                  code, inner_namespace=inner_namespace)
 
     def memorize_chunk(self, state, which_pass, data):
         for obj_name in state["pass_bins"][which_pass]:
-            self._eval(state["memorize_code"][obj_name], state, data)
+            self._eval(state["memorize_code"][obj_name], state["eval_env"],state, data)
```
njsmith: Missing space after comma, and probably a long line as well?

patsy/highlevel.py
```diff
@@ -33,7 +33,7 @@
 def _try_incr_builders(formula_like, data_iter_maker, eval_env,
                        NA_action):
     if isinstance(formula_like, DesignMatrixBuilder):
-        return (design_matrix_builders([[]], data_iter_maker, NA_action)[0],
+        return (design_matrix_builders([[]], data_iter_maker, eval_env, NA_action)[0],
```
njsmith: long line?

patsy/highlevel.py
```diff
@@ -53,7 +53,7 @@ def _try_incr_builders(formula_like, data_iter_maker, eval_env,
     if isinstance(formula_like, ModelDesc):
         return design_matrix_builders([formula_like.lhs_termlist,
                                        formula_like.rhs_termlist],
-                                      data_iter_maker,
+                                      data_iter_maker, eval_env,
```
njsmith: Another trivial nitpick: I prefer to either put all args on one line, or else put each arg on its own line -- easier to read that way IMO

njsmith:
Hi Christian! As you can see I've mostly recovered from PyCon, and did a quick pass for general style on what you have so far :-). Looking forward to seeing the rest!

patsy/build.py
```diff
@@ -614,7 +614,7 @@ def _make_term_column_builders(terms,
             term_to_column_builders[term] = column_builders
     return new_term_order, term_to_column_builders
 
-def design_matrix_builders(termlists, data_iter_maker, NA_action="drop"):
+def design_matrix_builders(termlists, data_iter_maker, eval_env=0, NA_action="drop"):
```
njsmith: After thinking about this a little more... I think maybe we should actually make `eval_env` a required argument here. This is the lower-level machinery for those who are doing clever programmatic things, so it's safer I think to insist the user think about how they want to handle the evaluation environment.

chrish42:
Hi Nathaniel! Thanks for the review. I will reply to them / apply them once I'm done with fixing the tests for using a subset of the environment. 

I am currently in the process of fixing said tests. Instead of comparing `state["eval_env"]` to the captured EvalEnvironment, I now build an EvalEnvironment containing the correct subset, and compare it to `state["eval_env"]`. However, that cannot work, as the comparison for equality of EvalEnvironments only compares on identity of each namespace dict. How would you feel if I made `EvalEnvironment.__eq__` check that each namespace dict has the same keys (by equality) and values (by identity) instead?

njsmith:
Hi Christian,

Sounds like a good plan. Before I answer your actual question, can you back
up a step and elaborate on why you need equality comparisons in the first
place? The old code needed to define equality on EvalEnvironment because it
had to define equality on EvalFactor, and EvalFactor used to have an
EvalEnvironment. But now that it doesn't, I think EvalEnvironment.__eq__
should just be unused (and maybe even removed)?

chrish42:
I need it for the tests. Here is one of the tests that I am in the process of fixing up (in eval.py):

```python
def test_EvalFactor_memorize_passes_needed():
    from patsy.state import stateful_transform
    foo = stateful_transform(lambda: "FOO-OBJ")
    bar = stateful_transform(lambda: "BAR-OBJ")
    quux = stateful_transform(lambda: "QUUX-OBJ")
    e = EvalFactor("foo(x) + bar(foo(y)) + quux(z, w)")

    state = {}
    eval_env = EvalEnvironment.capture(0)
    passes = e.memorize_passes_needed(state, eval_env)
    expected_eval_env = EvalEnvironment([{'foo': foo,
                                                  'bar': bar,
                                                  'quux': quux}],
                                                  eval_env.flags)
    assert passes == 2
    assert state["eval_env"] == expected_eval_env
# More stuff cut.
```

The last line is why I need equality on EvalEnvironment. (Before, the code just did `assert state["eval_env"] == eval_env`. Maybe my previous comment now makes more sense in light of this code? I think equality on EvalEnvironment needs to be defined (likely) as I described it for these tests. Although, if equality is used only for the tests, we could also pull it out of the class and put it into a test-only helper function? Anyways, let me know what you think is best here.

njsmith:
Ah, I see. My preference would still be to not define equality on `EvalEnvironment` at all (just because semantically it's extremely tricky -- e.g. two `EvalEnvironment` objects might act differently if they have different `dict` objects in them, because one thing you can do with an `EvalEnvironment` is execute code that modifies stuff in the environment -- and we don't really need it.) And for tests like this, just test that the `EvalEnvironment` does in fact have the correct behavior, e.g.:

```
# Variables in the current environment that were mentioned in the formula are accessible
assert state["eval_env"].eval("foo") is foo
# But other variables in the current environment that were not mentioned, are not available
assert_raises(NameError, state["eval_env"].eval, "state")
```

(This isn't as thorough as a full equality check, but we already have thorough tests for `EvalEnvironment.subset` and `ast_names`, right? So it isn't crucial to reproduce those tests here IMHO.)

chrish42:
Sounds reasonable to me. I'll move in that direction. I'll update the pull request when all the tests pass again.

chrish42:
Here's the updated pull request, using the subset of the environment in the state. Let me know what you think.

chrish42:
Also, what do you think is the right thing to do regarding shadowing of variables between the EvalEnvironment and the DataFrame or other passed to the formula?

patsy/desc.py
chrish42: It didn't percolate very far off the call stack, but I removed a couple uses. Good catch.

patsy/eval.py
chrish42: Good point. Done.

patsy/eval.py
chrish42: Done.

patsy/eval.py
chrish42: Added space, put each arg on its own line.

patsy/highlevel.py
chrish42: 86 chars, of which I've seen a couple other ones in the code base. Would you rather have me leave it as is, or split the arguments of the `design_matrix_builders()` function call each on its own line?

patsy/highlevel.py
chrish42: Oky doke. Done

patsy/build.py
chrish42: Sounds good to me too. Done.

patsy/build.py
chrish42: Yeah. I'm aware of this feature, but I vaguely remember being bitten by it (although maybe it was in C instead of Python), so I personally tend to prefer being explicit unless necessary. But given that this is the style of the rest of the code base, I'll go with it. Done.

patsy/highlevel.py
njsmith: Eh, whatever :-)

njsmith:
Looking pretty good! Things I can think of besides the shadowing issue:
* we need some additions to `doc/changes.rst` (esp. to document the incompatible changes!)
* need to do a quick pass through the docs to catch anything that's now out-of-date.
* I'd like to see an "end-to-end" test or two -- something that tests the user-visible behaviour directly, like running a formula through `dmatrix`, then modifying the original environment, and then running some more data through the same builder (like we were doing predictions) and checking that it is unaffected by our changes to the env. (test_highlevel.py is where the other tests like this go.)

njsmith:
Okay, so having had a chance to look over the code that would be affected by shadowing, I think my thought is: some sort of systematic handling of shadowing would probably be a good idea, but it feels like something we can just as well handle as an independent change. In particular, shadowing checks should probably also notice when builtins are shadowed (this has been a persistent issue -- see #13). So I don't think it's a blocker for this PR.

chrish42:
Working on the doc now. I assume you will register this release on zenodo before release time? (I'm also assuming the next release will be v0.4.0.)

njsmith:
Actually, an annoying thing about Zenodo is that you actually cannot get the DOI until after the release is finalized, basically b/c it the DOI-generation operation takes the release hash as an input. So I've just been putting the DOIs into the dev docs after the release, and the dev docs are what show up by default on readthedocs anyway, and I guess it's good enough.

And yeah, we'll call it v0.4.0.

chrish42:
Just updated the pull request with the results of my pass over the docs. Could you go over the diff for the doc update and see if things still make sense? I've deleted a lot more eval_env, etc. stuff than I added to the docs. I'm a bit worried that this means we missed an additional place where an eval_env argument should have been added.

I've also mentioned the incompatible change to `EvalFactor.memorize_passes_needed()` in the changelog, even though this is a pretty low-level method (which doesn't have a docstring too). Not too much detail for changes.rst? Anyways, feedback on my doc changes just as welcome as on the code ones.

I'll add the high-level tests another day, but other than that, this is starting to look pretty complete.

njsmith:
The doc changes look good to me -- I think the docs just spend more time talking about the factor interface than the actual builder interface, so it makes sense that you deleted more than you added. Mentioning the `EvalFactor.memorize_passes_needed` change is fine -- probably no-one cares, but oh well, it won't hurt.

chrish42:
Added a high-level test. Let me know if you have other ideas for these kinds of tests. Otherwise, please let me know if you have any final comments, if there are still missing bits, etc. (Really looking forward to getting this merged.)

njsmith:
Looks good to me!

One last thing: it looks like you have some merge conflicts which are greying out the merge button and preventing the tests from running. Could you merge or rebase from current master to resolve these?

chrish42:
Merge done.

njsmith:
Huzzah! We are good to go.

njsmith:
Thanks for the great work, and for sticking with it despite my nitpickiness :-).

chrish42:
Woot! Great!

chrish42:
Haha. No problem! Happy to be starting to make contributions on the scientific Python software stack. More of your comments were very relevant, I thought. And until I come close to you in terms of code contributed to patsy, it's much more your code base than mine, so some nitpickiness is fully allowed. (I tried already to stick to your coding style as best as I could.) I also much prefer code bases with a unified style anyway, even when I don't agree 100% with it. So no problem again.

Issue #61, opened 2015-04-14
Provide more metadata about how coding actually occurred, at the factor level


njsmith:
In `DesignInfo`, we should provide:

* For each factor, whether it is continuous or categorical
* If categorical, what levels it has
* If categorical, the `ContrastMatrix` we use for each (factor, term) pair.

This is pretty simple and obvious stuff, and useful to `statsmodels` (in particular, knowing what levels exist so you can generate all pairwise tests etc.).

njsmith:
CC @josef-pkt

Issue #26, opened 2013-10-06
Make DesignMatrixBuilders pickleable and saveable


njsmith:
Use cases:
- It should be possible to pickle a `DesignMatrixBuilder` (and/or `DesignInfo`, same issue)
- Checking if two designs are the same: this comes up for rERPy -- it's only valid to form a grand average across multiple analyses if the underlying regressions were the same. In particular it would be good to be able to check for subtle gotchas like use of `center(...)` with different means across the different analyses.

The easy part of this is reviewing the inner structure of `DesignMatrixBuilder` (column builders and all that) to make sure it's sensible, and similarly for factor state dicts.

The more complicated part is capturing the evaluation environment in a reasonable way.

Precondition: #25

Doron-Wiser:
Hey folks, 
The dill module (https://pypi.python.org/pypi/dill) is able to pickle DesignMatrixBuilders.
Just tried it!

njsmith:
Really? That's bad!

The problem isn't so much making it work at all, it's making it work in
such a way that a DesignMatrixBuilder pickled with patsy version X will
unpickle correctly on patsy version Y. If you use dill now it will probably
break later. (We should just have an error message, but I was lazy about
adding this because pickle didn't really work anyway.)

Also, the way dill "works" at the moment is basically to pickle the whole
universe just accidentally (b/c it will pickle every local variable that
happens to be sitting around in the environment where you used patsy, which
may well include your gigabyte sized dataset or whatever).

On Wed, Apr 15, 2015 at 3:25 PM, Doron-Wiser <notifications@github.com>
wrote:

> Hey folks,
> The dill module (https://pypi.python.org/pypi/dill) is able to pickle
> DesignMatrixBuilders.
> Just tried it!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pydata/patsy/issues/26#issuecomment-93538845>.
>



-- 
Nathaniel J. Smith -- http://vorpus.org

chrish42:
Yeah. Don't use dill to pickle v0.3 DesignMatrixBuilder objects. Patsy v0.4 will support pickling. (The harder bits are done.)

[...]

PR #67, opened 2015-05-21
Patsy pickling

chrish42:
Branch for the rest of the work required to make patsy pickles work well. For now, only a test that checks that builders roundtrip through pickle and give the same results afterwards. (The test already passes. We'll need to add at least another high-level test to check that design_info and builder objects make use the right pickle machinery to save only what's needed, etc.)

This will be a better place to discuss the rest of the changes required than the bug report (which notifies all other people with each comment). Let me know how you see the design for this.

- [ ] Figure out design.


njsmith:
Hey sorry, I'll get to this soon but I'll need to spend some time thinking hard about the current organization of `build.py` first, so it might be a few more days yet.

chrish42:
Any news?

njsmith:
Right, thanks for the ping :-).

I spent a few hours looking at this last night, and it's complicated!

The easy part: there are a bunch of simpler, lower-level objects that model pickling will rely on. For the most part I think we just need to add some tests for these (pickle/unpickle, and ideally having some historical pickles stored in the test suite to test that new versions of patsy continue to be able to unpickle them):
* `EvalFactor`
* `LookupFactor`
* `Term` (and in particular making sure its `.name()` method continues to work)
* `EvalEnvironment`
* all the stateful transform classes (in `state.py`, `splines.py`, `mgcv_cubic_splines.py`) -- there's some shared testing code for these in `test_state.py:check_stateful` that could take care of a lot of the work
* `ContrastMatrix`

Then there's the core design stuff, mostly in `build.py`, which currently includes:
* the two `FactorEvaluator` classes
* `_ColumnBuilder`
* `DesignMatrixBuilder`

This is the hard part, because the internal representations here are a mess -- weird encoding of information, redundant encodings across different objects, and so on. This really needs some substantive refactoring.

And then there's a few high-level objects that need some attention:
* `DesignInfo` -- a little complicated because it has a few different possible states -- sometimes only partial metadata is available -- but I think probably fine except for tests
* `DesignMatrix` -- currently inherits `ndarray.__reduce__`, which causes it to lose the crucial `.design_info` attribute when going through a pickle/unpickle cycle, so that needs fixing

So I'm going to keep thinking about the best way to handle `build.py`, but at least this should give some picture of the landscape...

njsmith:
It might also be worth doing a quick pass through the source to find any other classes that *aren't* on the above list, and add `def __getstate__(self): raise NotImplementedError` methods just to be safe...

njsmith:
Okay, I sat down and refactored big chunks of `build.py`, so now this should be easy :-). Basically I added a bunch more public metadata to `DesignInfo`, which is useful in its own right (#61), and then I redid stuff in `build.py` so that it can build matrices without ever using any non-public information. Which makes this much simpler.

I also went through and added raise-an-error `__getstate__` methods to everything.

So now I think the todo list is just:
* Go through all the classes mentioned above one at a time, removing the `__getstate__` method and adding pickle/unpickle tests.
* That's it.

chrish42:
That's great! I don't have time to look at this right now, but I should have more time around the next weekend.

njsmith:
Cool, no worries.

If I get some time I might go ahead and release 0.4.0, so people can at
least start taking advantage of the new stuff in master. But that's no big
deal, it would just mean the pickle stuff comes out in 0.5.0.

On Sun, Jun 14, 2015 at 5:03 PM, Christian Hudon <notifications@github.com>
wrote:

> That's great! I don't have time to look at this right now, but I should
> have more time around the next weekend.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pydata/patsy/pull/67#issuecomment-111888565>.
>



-- 
Nathaniel J. Smith -- http://vorpus.org

chrish42:
Working on this. Quick questions. Should I do a `__getstate__` implementation for everything, or is there some stuff that doesn't need to be pickled and can stay with `__getstate__ = no_pickling`? Also, what do you think would be good tests for this?

njsmith:
I guess we need a `__getstate__` (or `__reduce__` or whatever) for everything that's transitively referenced from `DesignInfo` or `DesignMatrix`. From the analysis in [my comment above](https://github.com/pydata/patsy/pull/67#issuecomment-110189598), I think that means: `EvalFactor`, `LookupFactor`, `Term`, `EvalEnvironment`, the stateful transform classes, `ContrastMatrix`, `DesignInfo`, `DesignMatrix`.

For tests, I'd generally suggest round-trip tests, plus some tests that involve stored pickle files (or strings inside the test .py, or whatever) that we attempt to load (since this is the only way to make sure we haven't accidentally broken both the save and load pathways in parallel, which is surprisingly easy to do with pickle).

chrish42:
Getting back to working on patsy. My plan is to write the pickling support and tests for a simple class (I picked `EvalFactor`) and show you that. Once we're happy with the way things are done for that class, I'll do the same approach for all the other classes.

So, while I'm in `EvalFactor`, I have a quick question. What's the `self.code` instance variable? It's not in the documentation of the class, and it also isn't used anywhere in said class. Do we still want to restore that on pickling?

Also, I assume we don't care about forward compatibility (i.e. it's okay if an older version of patsy can't read a pickle generated by a newer version), correct?

chrish42:
And for the tests, here's actually what I'm thinking of doing right now. You can tell me what you think, and if it sounds realistic.

We want to achieve two main things with the tests for pickling: 1) testing that pickling round-trips, and 2) testing that pickles created by old versions of patsy still work on newer version. I'd like not to repeat myself too much between these two, so here's what I'm thinking. I create a bunch of interesting patsy objects, all contained in (say) a dict. Then for each of these objects, I can test that `obj == pickle.loads(pickle.dumps(obj))`. I can also easily create a script that pickles all of these objects to disk, under a directory named after the version of patsy. That script would be run before each release. Then I can iterate over the directories for all the versions, and test that for each object, `obj == pickle.loads(the_pickle_for_that_object_for_a_given_version)`.

Do you think this would work as a testing strategy? Or am I missing something?

njsmith:
`EvalFactor.code` looks used to me?

Forward compatibility: It'd be good if trying to load a new pickle in an old version at least gave an error? But we can always deal with that later at the time we decide to change something in a breaking fashion. (I guess if one wants to get really fancy one could put a version number field in every pickle so that old versions will not just error out, but also be able to display a comprehensible error message - "This version of patsy is too old..." -- but this is definitely an extra credit kind of thing.)

The testing strategy also looks good to me. I guess the one subtlety will be that if we add new stuff in a future release then we'll need to have some way to add new test objects that are only present in newer pickle dumps, while still keeping the old pickle dumps.

One possible approach: represent the test objects as little bits of constructor code -- `"EvalFactor(...)"`, etc. -- and then have the version snapshot script dump both the pickles themselves and also the code used to create the objects that they should be compared to. (And then there can also be a test that loops over the same list for round-trips: `for code in TEST_OBJECT_CODES: obj = eval(code); assert pickle.loads(pickle.dumps(obj)) == obj`)

louispotok:
Hi! I just came across this PR via this issue: https://github.com/pydata/patsy/issues/26

Is this still the PR that would solve that issue? Is there a way that I can help here?

njsmith:
What's happened so far has happened here, yes :-). IIRC the general plan for how to handle this is clear, but stalled out with the work mostly not done. If you and @chrish42 want to work on it together, or if @chrish42 is busy and you want to just pick up his work and continue where he left off, then either way it would be great, and I'm happy to continue providing advice and review.

chrish42:
I've updated the description at the top to actually talk about the work that remains to be done, after going through all the comments in the pull request and other sources. @njsmith, let me know if that approach makes at least some sense or not. :-) After thinking about it some more, I still think that giving each "pickle test" its own name and corresponding object in a Python file is a better approach overall than saving the code as text inside a pickle. This mirrors other tests more (they all have a name too). We can discuss this approach more once I've added code to show how it would work in practice. (Will be easier to talk about it then.)

njsmith:
@chrish42: top level comment looks reasonable, and if you have a better idea I am extremely happy to defer to it :-)

chrish42:
This is not quite done or working yet, but it's starting to show the general approach.

chrish42:
@njsmith Now would be good time to have a look and give any comments about the general approach. If things look okay, we can start making more of the patsy objects pickeable, and add more pickling testcases. Comments and questions welcome!

patsy/test_pickling.py
```diff
@@ -0,0 +1,76 @@
...
+
+import six
+from six.moves import cPickle as pickle
+
+import os
+import shutil
+
+from patsy import EvalFactor
+
+PICKE_TESTCASES_ROOTDIR = os.path.join(os.path.dirname(__file__), '..', 'pickle_testcases')
```
njsmith: s/PICKE/PICKLE/, I assume?

patsy/test_pickling.py
```diff
@@ -0,0 +1,76 @@
...
+        # When adding features to a class, it will happen that there is no
+        # way to make an instance of that version version of that class
+        # equal to any instance of a previous version. How do we handle
+        # that?
+        # Maybe adding a minimum version requirement to each test?
+        assert pickling_testcases[testcase_name] == pickle.load(f)
+
+def test_unpickling_future_gives_sensible_error_msg():
+    # TODO How would we go about testing this?
+    pass
```
njsmith: I guess just test that we get a `PickleFromTheFuture` exception, or whatever it is we decide to call that?

patsy/test_pickling.py
```diff
@@ -0,0 +1,76 @@
...
+    finally:
+        os.rename(pickle_testcases_tempdir, pickle_testcases_dir)
+    print("Successfully created pickle testcases for new version {}.".format(version))
+
+if __name__ == "__main__":
+    import argparse
+
+    arg_parser = argparse.ArgumentParser(description="Create and save pickle testcases for a new version of patsy.")
+    arg_parser.add_argument("version", help="The version of patsy for which to save a new set of pickle testcases.")
+    args = arg_parser.parse_args()
```
njsmith: It might simplify the logic here if you know that our policy right now is that patsy version numbers are always of the form `x.y.z+dev` for development versions, and `x.y.z` only for released versions.

patsy/util.py
```diff
@@ -720,3 +720,23 @@ def test_safe_string_eq():
...
 
     assert not safe_string_eq(np.empty((2, 2)), "foo")
+
+def check_pickle_version(version, required_version, name=""):
+    if version > required_version:
+        error_msg = "This version of patsy is too old to load this pickle"
+    elif version < required_version:
+        error_msg = "This pickle is too old and not supported by this version of patsy anymore"
+    else:
+        return
```
njsmith: Presumably we'll eventually want to support multiple pickle versions at the same time -- but I guess we can always refactor this when the time comes :-)

njsmith: Specifically I think our goal will be to never stop supporting old pickles, so the "too old" error should never be needed... we'll see what happens though :-)

patsy/test_pickling.py
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+def test_pickling_same_version_roundtrips():
+    for obj in six.itervalues(pickling_testcases):
+        yield (check_pickling_same_version_roundtrips, obj)
+
+def check_pickling_same_version_roundtrips(obj):
+        assert obj == pickle.loads(pickle.dumps(obj, pickle.HIGHEST_PROTOCOL))
+
+def test_pickling_old_versions_still_work():
```
njsmith: I'd call this and the next function `unpickling_...` instead of `pickling_...`

patsy/test_pickling.py
njsmith: Also, hmm, so it looks like you're storing the pickle tests outside the source tree. Right now we go to some effort to ship the test suite, so that we can test patsy-as-installed... and I think the current CI stuff tries hard to run the test suite without access to the source tree, to make sure that this works. So this is definitely a change. Maybe it's a good change if the pickle test suite is large... but we'll need to do some more work in that case to make it really work. OTOH, maybe the pickle test suite is small, in which case we might as well ship it, because testing is important.

Probably the simplest thing to do is to move it into a `patsy/test_data` directory inside the package for now, and then we can always move it back out again if we decide that we have too many megabytes of test pickles. (This will probably require some annoying change to setup.py as well to include the test data -- [example](https://github.com/njsmith/zs/blob/42ca7679c2b986243abcff048bc42d49c204048c/setup.py#L58-L77).)

patsy/test_pickling.py
```diff
@@ -0,0 +1,76 @@
...
+            if os.path.splitext(fname)[1] == '.pickle':
+                yield check_pickling_old_versions_still_work, os.path.join(dirpath, fname)
+
+def check_pickling_old_versions_still_work(pickle_filename):
+    testcase_name = os.path.splitext(os.path.basename(pickle_filename))[0]
+    with open(pickle_filename, 'rb') as f:
+        # When adding features to a class, it will happen that there is no
+        # way to make an instance of that version version of that class
+        # equal to any instance of a previous version. How do we handle
+        # that?
```
njsmith: typo: `version version`

And I think the answer here is probably that when we unpickle an old object into a new patsy, if the new patsy has some new features, then we'll have to somehow initialize them for the old object (like fill in zeros or empty strings or whatever), because we need *some* way to represent it in memory. So we can put in a test case here that uses those default field values, and also add a new object that actually uses the new fields in a non-trivial way.

njsmith:
I made a bunch of small comments, but the overall approach looks fine to me! Thanks for working on this, and sorry for being slow in reviewing.

PR #74, opened 2015-10-08
[doc] How to access the EvalEnviroment namespace

has2k1:
Fixes the example as show [here](http://patsy.readthedocs.org/en/latest/API-reference.html#patsy.EvalEnvironment.capture).


njsmith:
Looks good, thanks!

I'm curious now what you're doing that requires touching `EvalEnvironment` directly :-)

has2k1:
Using it in ggplot.

PR #75, opened 2015-10-18
Unicode message

has2k1:
The message baffled me way longer than it should have. Cherry-picked and modified the original solution (#56).

### DIFF
--- a/patsy/highlevel.py
+++ b/patsy/highlevel.py
@@ -13,6 +13,7 @@ __all__ = ["dmatrix", "dmatrices",
 #   ModelDesign doesn't work -- need to work with the builder set
 #   want to be able to return either a matrix or a pandas dataframe
 
+import six
 import numpy as np
 from patsy import PatsyError
 from patsy.design_info import DesignMatrix, DesignInfo
@@ -56,6 +57,9 @@ def _try_incr_builders(formula_like, data_iter_maker, eval_env,
                                       eval_env,
                                       NA_action)
     else:
+        # check if formula_like is another string type (which might be unsupported)
+        if isinstance(formula_like, string_types):
+            raise PatsyError("Unsupported string type for formula (e.g., unicode in Python 2.X).")
         return None
 
 def incr_dbuilder(formula_like, data_iter_maker, eval_env=0, NA_action="drop"):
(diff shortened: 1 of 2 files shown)


patsy/highlevel.py
```diff
@@ -55,8 +56,8 @@ def _try_incr_builders(formula_like, data_iter_maker, eval_env,
                                       data_iter_maker,
                                       eval_env,
                                       NA_action)
-    else:
-        return None
+    if not six.PY3 and isinstance(formula_like, unicode):
```
JanSchulz: Is there a way to first try to cast to str and check if both are the same?

```
a = u"ABC"
b = u"ÄAÜ"
for x in [a,b]:
    print(x.encode('ascii','replace') == x)
# prints:
True
False
```

has2k1: I think that is better than just bailing out.

```python
if not six.PY3:
    try:
        formula_like = str(formula_like)
    except UnicodeEncodeError:
        raise PatsyError('...')
```

njsmith:
If the goal is to auto-cast when possible, then the right way to do that is
```
if not six.PY3 and isinstance(obj, unicode):
    try:
        obj = obj.encode("ascii")
    except UnicodeEncodeError:
        raise PatsyError(...)
```
and it might even go into `ModelDesc.from_formula`. But... while issuing a better error message would obviously be a win, I'm less sure that auto-casting is actually a good idea. Can someone elaborate on the motivation for this? My intuition is that it's not really a convenience for interactive use, because in interactive use no-one is going to be typing `u"y ~ x"` when they could be typing `"y ~ x"` -- so I assume these unicode strings are coming from some other library/framework. And if you have some code that is dealing with unicode strings (esp. on py2), then your only hope of making your code correct is if you have an explicit strategy for dealing with unicode/encoding issues. Should patsy really be making a wild guess at what to do when passed unicode? In what case is this better than raising an error and making the caller think about what they actually want (which may or may not be `.encode("ascii")`)? (This is a real question.)

has2k1:
The main culprit is something of the sorts

```python
from __future__ import ..., unicode_literals, ...
```
It is supposed to be a magic bullet to unicode on py2, but it malfunctions in weird ways.

If there is any uncertainty about casting then it should not be done, a helpful error message is enough.

njsmith:
Okay, `from __future__ import unicode_literals` convinces me :-) I put together a more complete version in #76 -- I'll close this, and would appreciate it if you folks could review that.

Issue #53, opened 2015-01-31
Providing a unicode string to dmatrices results in a misleading error message


gnovak:
Doing this works fine:
foo = pd.DataFrame(dict(a=[1,2.0], b=[4,5.0]))
patsy.dmatrices('a ~ b', foo)

However, if the formula is given by a unicode string:
patsy.dmatrices(u'a ~ b', foo)

An exception is raised:
ValueError: design matrix must be real-valued floating point

The exception is very misleading about the nature of the problem.  I thought an inf or nan had crept into the dataframe, but no bad values were there.

One possible fix is to send the first argument of dmatrices through the str() function:
patsy.dmatrices(str(u'a ~ b'), foo)

Another possible fix is to raise print a different error message so that the user has some clear idea of the cause of the problem. 

I'm using patsy version 0.3.0

Thanks!

cjl2183:
thank you! this helped me. agreed that the error msg is extremely misleading.

PR #76, opened 2015-10-27
Allow unicode formulas on Py2 in limited circumstances.

njsmith:
Apparently our disallowing unicode formula strings on Py2 was being
rather annoying for people using `from __future__ import
unicode_literals`. Start allowing them in limited circumstances.

Fixes gh-53.


has2k1:
It looks good.

Issue #82, opened 2016-04-06
Can't seem to get scaled diff's


rsgmon:
`In[92]: df = pd.DataFrame([(1,3),(2,6),(4,2),(6,5),(7,3),(4,6),(2,2),(6,4)], columns=['y','X'])`
`In[93]: pt.dmatrices('y ~ X.diff()', df)`
Out[93]: 
(DesignMatrix with shape (7, 1)
   y
   2
   4
   6
   7
   4
   2
   6
   Terms:
     'y' (column 0),
 DesignMatrix with shape (7, 2)
   Intercept  X.diff()
           1         3
           1        -4
           1         3
           1        -2
           1         3
           1        -4
           1         2
   Terms:
     'Intercept' (column 0)
     'X.diff()' (column 1))

`In[94]: pt.dmatrices('y ~ scale(X.diff())', df)`

Traceback (most recent call last):
  File "C:\Python34\lib\site-packages\IPython\core\formatters.py", line 222, in catch_format_error
    r = method(self, *args, **kwargs)
  File "C:\Python34\lib\site-packages\IPython\core\formatters.py", line 699, in __call__
    printer.pretty(obj)
  File "C:\Python34\lib\site-packages\IPython\lib\pretty.py", line 368, in pretty
    return self.type_pprinters[cls](obj, self, cycle)
  File "C:\Python34\lib\site-packages\IPython\lib\pretty.py", line 552, in inner
    p.pretty(x)
  File "C:\Python34\lib\site-packages\IPython\lib\pretty.py", line 382, in pretty
    return meth(obj, self, cycle)
  File "C:\Python34\lib\site-packages\patsy\design_info.py", line 1089, in _repr_pretty_
    for col in formatted_cols]
  File "C:\Python34\lib\site-packages\patsy\design_info.py", line 1089, in <listcomp>
    for col in formatted_cols]
ValueError: max() arg is an empty sequence
Out[94]:

njsmith:
FYI -- to paste multi-line code blocks on github, use triple-backquotes. (I just fixed your original post -- if you click "edit" on it you can see how I modified it.)

The main problem you are hitting here is that your `X.diff()` thing has a NaN in it:
```python
In [11]: df["X"].diff()
Out[11]: 
0   NaN
1     3
2    -4
3     3
4    -2
5     3
6    -4
7     2
Name: X, dtype: float64
```
[...]

PR #83, opened 2016-04-06
Fix a crash in DesignMatrix.__repr__ when shape[0] == 0

njsmith:
Discovered by @rsgmon in gh-82.

### DIFF
--- a/patsy/design_info.py
+++ b/patsy/design_info.py
@@ -1085,8 +1085,13 @@ class DesignMatrix(np.ndarray):
             formatted_cols = [_format_float_column(PRECISION,
                                                    printable_part[:, i])
                               for i in range(self.shape[1])]
-            column_num_widths = [max([len(s) for s in col])
-                                 for col in formatted_cols]
+            def max_width(col):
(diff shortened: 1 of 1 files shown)


njsmith:
Test failures are because doing `conda install pandas=0.14.0` apparently produces a totally broken conda environment (`import numpy` -> error: libgfortran not found; see Travis logs). Odd. I don't think we should stop testing patsy on pandas 0.14.0 yet -- at least not without ripping out all the old support code for it! -- but I don't want to deal with this right now, so I guess I'll merge and worry about it later.

@jreback: I feel like I should bring this weird brokenness in anaconda to someone's attention, but I'm not sure who, and you hit my intersection of pandas + continuum + conda advocate :-)

jreback:
yeah IIRC there was an issue with some meta data in the dependencies in older numpies, you can make this work by forcing

``libfortran=1.0``

e.g. ``conda install -n pt numpy==1.8.2 libfortran==1.0`` (only needed for linux)

jreback:
https://github.com/conda/conda/issues/444

PR #91, opened 2016-09-06
xarray support for categorical data

ceholden:
Not sure if this is out of scope for the project, but this PR tries to smooth over the only issue I've came across so far when using Patsy with `xarray.Dataset` objects (in place of `pd.DataFrame`).

As a bit of background, xarray extends the idea of the data frame into an arbitrary number of dimensions, is now integrated into pandas, and is being considered as a replacement for the Panel capabilities in pandas. Xarray has two main data types -- `DataArray` and `Dataset` -- that are n-dimensional extensions of Panda's `Series` and `DataFrame` (see the [docs here](http://xarray.pydata.org/en/stable/why-xarray.html#core-data-structures)). I love using Patsy in my work with geospatial data, so such an integration would be great for me and might help down the road if Pandas deprecates Panel in favor of xarray objects.

To my surprise, feeding `xarray.Dataset` objects containing purely numeric data as the data into Patsy design matrix calls worked without any hangup. However, the integration breaks down when one tries to create a design matrix involving categorical data:

``` python
patsy.PatsyError: Error interpreting categorical data: all items must be hashable
```

For example, this error crops up in [`categorical.CategoricalSniffer:sniff`](https://github.com/pydata/patsy/blob/8b6c7121e7ec3f95da1fd9b9760945835abff8b3/patsy/categorical.py#L211). It seems the root of the issue is that the `__iter__` for an `xarray.DataArray` returns a single element still wrapped up in an `xarray.DataArray`, unlike when iterating over a `pd.Series` or `np.ndarray`.

I gave a go at a patch by defining `have_xarray` in `patsy/utils` and "unwrapping" data from its `xarray.DataArray` container before trying to perform any of the checks or conversions Patsy does with categorical data. I'm happy to iterate on design or implementation if you think it's worth integrating.

Thanks for your work!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ before_install:
   - conda info -a
   - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION numpy scipy coverage nose pip
   - source activate testenv
-  - if [ "$PANDAS_VERSION_STR" != "NONE" ]; then conda install pandas${PANDAS_VERSION_STR}; fi
+  - if [ "$PANDAS_VERSION_STR" != "NONE" ]; then conda install pandas${PANDAS_VERSION_STR} xarray; fi
 install:
   - python setup.py sdist
   - pip install dist/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,9 @@ matrix:
       env: PANDAS_VERSION_STR="=0.14.0 libgfortran=1.0"
     # 0.18.0 has is_categorical_dtype in a different place than 0.19.0+
     - python: 3.4
-      env: PANDAS_VERSION_STR="=0.18.0"
+      env: PANDAS_VERSION_STR="=0.18.0" XARRAY_VERSION_STR=">=0.7.0"
     - python: 2.7
-      env: PANDAS_VERSION_STR="=0.18.0"
+      env: PANDAS_VERSION_STR="=0.18.0" XARRAY_VERSION_STR=">=0.7.0"
     # make sure it works without pandas
     - python: 3.5
       env: PANDAS_VERSION_STR="NONE"
@@ -41,7 +41,8 @@ before_install:
   - conda info -a
   - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION numpy scipy coverage nose pip
   - source activate testenv
-  - if [ "$PANDAS_VERSION_STR" != "NONE" ]; then conda install pandas${PANDAS_VERSION_STR} xarray; fi
+  - if [ "$PANDAS_VERSION_STR" != "NONE" ]; then conda install pandas${PANDAS_VERSION_STR}; fi
+  - if [ -n "$XARRAY_VERSION_STR" ]; then conda install xarray${XARRAY_VERSION_STR}; fi
 install:
   - python setup.py sdist
   - pip install dist/*
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -236,6 +236,8 @@ intersphinx_mapping = {"python": ("http://docs.python.org", None),
                                  None),
                        "pandas": ('http://pandas.pydata.org/pandas-docs/stable/',
                                   None),
+                       "xarray": ('http://http://xarray.pydata.org/en/stable/',
+                                  None),
                        }
 
 autodoc_member_order = "source"
--- a/doc/quickstart.rst
+++ b/doc/quickstart.rst
@@ -24,8 +24,9 @@ variables:
 Of course Patsy doesn't much care what sort of object you store
 your data in, so long as it can be indexed like a Python dictionary,
 ``data[varname]``. You may prefer to store your data in a `pandas
-<http://pandas.pydata.org>`_ DataFrame, or a numpy
-record array... whatever makes you happy.
+<http://pandas.pydata.org>`_ DataFrame, an
+`xarray <http://xarray.pydata.org>`_ Dataset (using variables that have
+1 or 2 dimensions), or a numpy record array... whatever makes you happy.
 
 Now, let's generate design matrices suitable for regressing ``y`` onto
 ``x1`` and ``x2``.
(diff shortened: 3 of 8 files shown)


njsmith:
This seems reasonable at a quick skim, but because the travis build isn't installing xarray, none of the new code is actually getting tested. (You can also see this in the coverage bots complaining about lots of untested lines.) Could you add xarray to the list of libraries in .travis.yml?

ceholden:
Thanks for the review. I added in `xarray` to the Travis install list for compatible versions of Pandas as it made sense to me (2nd commit failed on tests with `pandas=0.14`), but I'm happy to edit to accommodate any preferences you might have.

shoyer:
> I added in xarray to the Travis install list for compatible versions of Pandas as it made sense to me (2nd commit failed on tests with pandas=0.14)

xarray requires pandas 0.15 or newer, so indeed that makes sense to me.

shoyer:
Does this only work on 1D `xarray.Dataset` objects?

It might also be nice to mention xarray compatibility in the patsy docs somewhere (e.g., in the docstring for `dmatrix`).

ceholden:
Good point about mentioning the support in the docs. I've added some text to the `dmatrix` docstring and quickstart page as suggested.

Technically it will work for 1- or 2-dimensional data, just like Patsy works with 1- or 2-dimensional NumPy arrays or DataFrames. I say "technically" because I didn't know it worked with 2-dimensional data until just now and I'm not really sure what the use case might be. So, this works for example:

``` python
> import numpy as np
> import patsy
> import xarray as xr
> shp = (10, 20)
> ds = xr.Dataset({'a': (['time', 'x'], np.random.rand(*shp))},
                  coords={'time': np.arange(shp[0]),
                          'x': np.arange(shp[1])})
> patsy.dmatrix('1 + a + time', ds)
DesignMatrix with shape (10, 22)
  Columns:
    ['Intercept',
     'a[0]',
     'a[1]',
...
     'a[19]',
     'time']
  Terms:
    'Intercept' (column 0), 'a' (columns 1:21), 'time' (column 21)
  (to view full data, use np.asarray(this_obj))
```

Passing more dimensions than the 2 dimensions allowed triggers [a PatsyError from this function](https://github.com/pydata/patsy/blob/a5b54c213b93f7dc23008ff44a3de66285703e95/patsy/build.py#L42) regardless of if you use a NumPy array or `xarray` object, so I don't think we need special handling for the `xarray` case. 

In case you're interested: my use cases the data is usually only along a single `time` coordinate with space coordinate squeezed out with a `ds.sel(x=x, y=y)` sort of call. I use `xarray` to organize, align, and store time series of remote sensing data, sometimes from different sensors with different pixel resolutions (hooray for `reindex_like(other, method='nearest')`). I might do some calculations with the `xarray` object, like derive a "spectral index" that contrasts some of the variables, but the intersection with `patsy` comes in when trying to statistically model each pixel as an individual. I don't really ever put design matrices back into an `xarray` object so I wasn't interested in having it supported as a return type.

shoyer:
I've used patsy in the past with xarray, but mostly just by converting into a `pandas.DataFrame` first (via `.to_dataframe`). There are then various ways you can pack stuff back into an xarray object.

In principle, it would be nice if patsy had generic support for handling multi-dimensional `xarray.Dataset` -> `xarray.Dataset` with new variable names. Could probably make this work (with a few extraneous memory copies) by stacking the dataset with `.stack()` to make it 1D, passing a dict of 1D numpy arrays into `patsy.DesignMatrix`, rebuilding an `xarray.Dataset` with the 1D coordinates and then unstacking back into the original dimensions.

njsmith:
The general rule is that patsy is happy to accept 1d or 2d numerical predictors, but only 1d categorical predictors, mostly because I don't know what use a 2d categorical predictor would be. (Maybe someone will tell me.)

Though I guess for most uses of multi-dimensional xarrays, the multiple dimensions represent something like lat x lon, whereas patsy always works in observations x predictors space which is totally different. If you do give patsy a 2d ndarray, then it interprets it as observations x predictors -- e.g. something like `poly(x)` returns a matrix where rows correspond to entries in `x` while columns are the different polynomial orders. So it might make sense to have utilities for flattening xarrays into something like observations x predictors space... I'm not sure why it would be useful to reshape back to xarrays on the output, though? It seems like the next step would be to feed into something like a least squares solver, which also works in observations x predictors space.

njsmith:
Does this code handle all of these cases?

* `"1 + x"` where `x` is in an xarray and contains numerical values
* `"1 + x"` where `x` is in an xarray and contains categorical values
* `"1 + C(x)"` where `x` is in an xarray and contains numerical values (e.g. integers)
* `"1 + C(x)"` where `x` is in an xarray and contains categorical values

shoyer:
> So it might make sense to have utilities for flattening xarrays into something like observations x predictors space... I'm not sure why it would be useful to reshape back to xarrays on the output, though? It seems like the next step would be to feed into something like a least squares solver, which also works in observations x predictors space.

The use case would be for fitting, e.g., a collection of linear models. For example, you have a number of raw features (e.g., physical variables) defined on a grid (latitude, longitude, time). You combine these raw features with patsy to build predictors, also on the same grid. Then you a bunch of independent time series models on (time, feature) for each (latitude, longitude) grid point.

This sort of analysis is pretty common in the geosciences, e.g., to understand climate trends.

Anyways, this is definitely beyond the scope of this PR!

ceholden:
@shoyer has basically described my exact use case, though I think for a different set of science questions and datasets in mind :). The advantage of the change he's described seems to be mostly a better user experience and it'd probably be faster to stack and calculate all in one go instead of running the same design matrix calculation inside of a "select pixel at lon/lat". Something to think about adding via another PR if Patsy is amenable to supporting this use.

@njsmith I've updated the test suite to test categorical and numeric data with and without using `C()` in the formula. Should be a "yes" to your question.

PR #97, opened 2016-10-24
ENH/BUG: Let patsy use categoricals when they exist... BIG speed improvement

thequackdaddy:
So when a column is declared a categorical, patsy is skipping currently skipping all the `pandas` builtins that greatly speed up the process. 

I think there was a piece of the code out-of-order here:

https://github.com/pydata/patsy/blob/master/patsy/categorical.py#L312

I think we want to set `data = data.data` earlier to take advantage of pandas.

So before making this change:

``` python
from patsy import dmatrix
import numpy as np
import pandas as pd

x = np.random.choice(list('abcdefg'), size=1e7)
x = pd.Series(x, dtype=pd.Categorical)


C:\Program Files\Anaconda3\lib\site-packages\spyder\utils\ipython\start_kernel.py:5: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  # (see spyder/__init__.py for details)

%time dmatrix('C(x)')


Wall time: 18.7 s
```

And after..

``` python
from patsy import dmatrix
import numpy as np
import pandas as pd

x = np.random.choice(list('abcdefg'), size=1e7)
x = pd.Series(x, dtype=pd.Categorical)


C:\Program Files\Anaconda3\lib\site-packages\spyder\utils\ipython\start_kernel.py:5: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  # (see spyder/__init__.py for details)

%time dmatrix('C(x)')


Wall time: 1.14 s
```

### DIFF
--- a/patsy/categorical.py
+++ b/patsy/categorical.py
@@ -42,6 +42,7 @@ from patsy.util import (SortAnythingKey,
                         have_pandas, have_pandas_categorical,
                         have_pandas_categorical_dtype,
                         safe_is_pandas_categorical,
+                        safe_pandas_Categorical_reorder,
                         pandas_Categorical_from_codes,
                         pandas_Categorical_categories,
                         pandas_Categorical_codes,
@@ -322,9 +323,9 @@ def categorical_to_int(data, levels, NA_action, origin=None):
                              % (levels, data_levels_tuple), origin)
         if not data_levels_tuple == levels:
             if isinstance(data, pandas.Categorical):
-                data = data.reorder_categories(levels, ordered=False)
+                data = safe_pandas_Categorical_reorder(data, levels)
             else:
-                data = data.cat.reorder_categories(levels, ordered=False)
+                data = safe_pandas_Categorical_reorder(data.cat, levels)
         # pandas.Categorical also uses -1 to indicate NA, and we don't try to
         # second-guess its NA detection, so we can just pass it back.
         return pandas_Categorical_codes(data)
--- a/patsy/categorical.py
+++ b/patsy/categorical.py
@@ -309,20 +309,27 @@ def categorical_to_int(data, levels, NA_action, origin=None):
     assert isinstance(levels, tuple)
     # In this function, missing values are always mapped to -1
 
-    if safe_is_pandas_categorical(data):
-        data_levels_tuple = tuple(pandas_Categorical_categories(data))
-        if not data_levels_tuple == levels:
-            raise PatsyError("mismatching levels: expected %r, got %r"
-                             % (levels, data_levels_tuple), origin)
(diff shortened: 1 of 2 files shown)


njsmith:
Sorry, I'm having trouble figuring out what this change is doing. There's something about moving a check earlier, and it's mixed in with a behavioral change to level ordering (but maybe just for pandas categoricals), and somehow things are faster...? Or something like that? Can you explain it to me like I'm 5? :-)

thequackdaddy:
@njsmith Sorry for lacking detail. I'm quite confused by it myself! Trust me, the developers of this tool are much more competent programmers than I am. I sort of only accidentally stumbled onto this behavior. 

Try this code...

```python
In [8]: import numpy as np
   ...: import pandas as pd
   ...: from patsy import dmatrix
   ...: 
   ...: x = np.random.choice(list('abcdefg'), 10000000)
   ...: x = pd.Categorical(x)
   ...: 

In [9]: %time dmatrix('x')
Wall time: 2.88 s

In [10]: %time dmatrix('C(x)')
Wall time: 46.5 s
```
So something strange is happening when you use the `C(x)` command on a categorical. 

I think the problem is here...

https://github.com/pydata/patsy/blob/master/patsy/categorical.py#L308

In this function when you put any factor in the `C` function, it enters the function as a `_CategoricalBox`.  The categorical box stores the data as `data.data`, so you need to un-box the categorical stuff by doing a `data = data.data` before checking if `data` is safe to use as a pandas categorical. The safe pandas check will return False when a `_CategoricalBox` is sent. 

Since the pandas categorical fails, it has to go through the whole `NAAction` check thing, which is super-comprehensive (yay!) yet not as fast as I would like it. 

So what this PR does is first un-box the Categorical box, then check if its categorical. 

The re-ordering piece was just because some tests were failing because patsy checks if the specified order levels in the `C` function match the order of the categories. pandas Series that use a string object type can be re-ordered, and the tests (AFAICT) assume that categoricals can be re-ordered too. 

However, there is another test that is supposed to raise an error when a Categorical variable has levels in a different order than the categories... To make those tests pass, I use the `reorder_categories` method. That didn't exist prior to pandas 0.15.0 (I think), so for pandas 0.14.0 to pass, I reorder by hand (which would be very slow). 

You could reject this PR and just tell users that if you are using a categorical to not use the `C` function. However, this allows fully-functioning `Helmert` and `Poly` coding at much, much faster speeds. 

And personally, I've always used `C(x)` coding in R, even on factors that were already categories... Just a nice reminder of what I'm doing to that factor. So this lets you continue to use that practice and still be really fast. 

Phew. That was a mouthful.

Issue #93, opened 2016-09-16
patsy question/wishlist


thequackdaddy:
Hello,

I fit some (relatively) large-ish GLMs in statsmodels and have been experimenting with using `patsy` instead of a home rolled thing. My home rolled method isn't very good (I tend to underestimate challenges...). I've gotten some better hardware so now some models that used to not work with patsy (because of memory constraints) work now. I've run across a few things that might make it easier for me to use patsy more. Happy to work on PRs for them if there's interest. 

1. Categorical NA logic: Currently, it appears that when a categorical is fed through `patsy`, every inidividual value is checked against a rather detailed list on how to handle NaNs/Missing Values/Empty whatever. I ran a cProfile on this and it was quite slow. I think the bottleneck is here:
https://github.com/pydata/patsy/blob/master/patsy/categorical.py#L341
I know that NaNs/NA/None/Empty is a mess in general, but its a fact of life in my line of work (insurance modeling). I'm wondering if we scope out exactly all the scenarios we need to control for and use (pandas maybe?) to do this more elegantly? I'm not sure of the scope as there's far more players here than me.

2. Reading from on-disk data stores: Memory is something of a problem for me. a typical model that I run might eat up 10+ GB of RAM. It works, but obviously is not ideal. As far as (relatively) mature tools, I've found bcolz's ctables to be pretty good (and fast). HDFStores/dask would be nice too. I'm not sure if #91 relates to this (I don't know if xarray works well as an on-disk storage/data tool). 

3. Partial predictions: This is partly a statsmodels partly a patsy idea... I'd like the ability to do a so-called partial predict. Essentially I have a model like `y ~ a + b + a:c`. I want to come up with predictions `y` assuming that just `a` changes or just `b` changes. I think the process would look something like (assuming we're talking about changing only `a`) creating a new design matrix with every unique value of `a` as a separate row, and have the most frequent (or some other innocuous value) of `b` and `c` as constant for these rows. Then feed this dataset through the statsmodels `predict` routine. This is very helpful for GLMs with the log link--which is the bulk of what I work with.

thequackdaddy:
For item 1 here, #97 greatly improves categorical coding time... from 10's of seconds on million record datasets to only a few seconds. 

NA/NaN/NaT/None and the like are a real nightmare...

njsmith:
> Reading from on-disk data stores: Memory is something of a problem for me. a typical model that I run might eat up 10+ GB of RAM

[...]

PR #98, opened 2016-12-29
ENH: Generate var_names from the data

thequackdaddy:
Hello,

I have a proposal that really came about because of the way I've been interacting with patsy. 

My datasets are kind of long and kind of wide. I have lots of fields that I use for expoloring stuff, but naturally they just don't work out. 

I've been using bcolz because it stores the data in a columnar fashion making horizontal slices really easy. Before, I'd been creating a list of variables that I wanted, defining all the transforms that I needed in patsy, and then feeding that through. I can't load the entire dataset into memory just because its too wide and long and I might only be looking at 20-30 columns for any one model. 

So I propose having patsy attempt to figure out which columns it needs from the data using this new `var_names` method which is available on `DesignInfo`, `FactorInfo`, and `Term`. In a nutshell, it gets a list of all the variables used, checks if that variable is defined in the `EvalEnvironment`, and if not, assumes it must be data. 

I've called this `var_names` for now, but arguably maybe `non_eval_var_names` might be more accurate? Open to suggestions here. 

One nice thing is that when using `incr_dbuilder`, it can automatically slice on the columns which makes the construction much faster (for me at least). 

Here's a gist demo'ing this. 

https://gist.github.com/thequackdaddy/2e601afff4fbbfe42ed31a9b2925967d

Let me know what you think.


thequackdaddy:
I went ahead and built the `partial` function that I had alluded to in #93. This makes it much easier to create design matrices for `statsmodels` that show you the marginal differences whe you only change the levels of 1 (or more) factors. 

Here's a basic example:

```python
In [1]: from patsy import dmatrix
   ...: import pandas as pd
   ...: import numpy as np
   ...:
   ...: data = pd.DataFrame({'categorical': ['a', 'b', 'c', 'b', 'a'],
   ...:                      'integer': [1, 3, 7, 2, 1],
   ...:                      'flt': [1.5, 0.0, 3.2, 4.2, 0.7]})
   ...: dm = dmatrix('categorical * np.log(integer) + bs(flt, df=3, degree=3)',
   ...:  data)
   ...: dm.design_info.partial({'categorical': ['a', 'b', 'c']})
   ...:
Out[1]:
array([[ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  1.,  0.,  0.,  0.,  0.,  0.,  0.,  0.],
       [ 0.,  0.,  1.,  0.,  0.,  0.,  0.,  0.,  0.]])

In [2]: dm.design_info.partial({'categorical': ['a', 'b'],
   ...:                         'integer': [1, 2, 3, 4]},
   ...:                        product=True)
Out[2]:
array([[ 0.        ,  0.        ,  0.        ,  0.        ,  0.        ,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  0.69314718,  0.        ,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  1.09861229,  0.        ,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  0.        ,  0.        ,  1.38629436,  0.        ,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  1.        ,  0.        ,  0.        ,  0.        ,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  1.        ,  0.        ,  0.69314718,  0.69314718,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  1.        ,  0.        ,  1.09861229,  1.09861229,
         0.        ,  0.        ,  0.        ,  0.        ],
       [ 0.        ,  1.        ,  0.        ,  1.38629436,  1.38629436,
         0.        ,  0.        ,  0.        ,  0.        ]])
```

thequackdaddy:
@njsmith Also, it appears that travis isn't kicking off for this all of a sudden. Any ideas why this would be?

njsmith:
It seems like it would be simpler to query a `ModelDesc` for all the variables it uses, period? And then it'd be your job to ignore the ones that aren't present in your data set. This would also be more accurate, because 

The even simpler (from the API design / coupling perspective) would be to pass in a dict-like that loads the data lazily on demand, like:
```python
class LazyData(dict):
    def __missing__(self, key):
        try:
            return bcolz.load(key, file)
        except BcolzKeyNotFound:
            raise KeyError(key)
```
Would this work for you?

Is the `partial` part somehow tied to the `var_names` part? They look like separate changes to me, so should be in separate PRs?

This is also missing lots of tests, but let's not worry about that until after the high-level discussion...

thequackdaddy:
> It seems like it would be simpler to query a ModelDesc for all the variables it uses, period? And then it'd be your job to ignore the ones that aren't present in your data set. This would also be more accurate, because

Hmm... I hadn't thought of that. That should be relatively easy to add/change based on what I've done so far. The heart of this is the `var_names` is on the `EvalFactor` class that looks at all the objects needed to evalulate the factor using the `ast_names` function. This is in turn used by the `Term` class... (and is used in turn by `DesignInfo` class). `ModelDesc` has a list of terms (`lhs_termlist` and `rhs_termlist`), so adding this would be easy. 

I presume you're implying that I shouldn't be worrying about the `EvalEnvironment` variable and just return every dependent object--function and module alike? I was trying to return only "data"-ish things. Simply removing them from the output set manually set seems easy enough... 

 > The even simpler (from the API design / coupling perspective) would be to pass in a dict-like that loads the data lazily on demand, like:

This is really clever, thanks! I'll try it. However, I don't think it solves the `partial` issue below. 

> Is the partial part somehow tied to the var_names part? They look like separate changes to me, so should be in separate PRs?

Yes. `partial` looks at each `Term`'s `var_names` and decides whether the `Term` needs the variable or not. If yes, it pulls that `Term` using `subset` to create the design matrix only for that subset of columns using the variables specified. Otherwise, it returns columns full of zeros. The end result is a design matrix of the same width and column alignment  as the model's `DesignMatrix`, but only with as many rows as needed to evaluate the partial predictions and the rest of the columns as zeros. 

> This is also missing lots of tests, but let's not worry about that until after the high-level discussion...

Sound good. Writing tests is not something I've excelled at. This is somewhat tested and I (think) there is coverage for most of the new lines--likely I missed a few. I added some `asserts` to some of the existing tests with the new functionality.

PR #104, opened 2017-04-06
Pickles 

thequackdaddy:
Hello,

This is a continuation effort of #86 and #67. My apologies if this is viewed as stepping on someone else's toes... but this is an interesting--and useful topic--for me. 

I've basically got this working (with some simple tests on real data), but some things break some of the unit tests. 

One problem is that as far as I can tell, doing a `pickle.loads(pickle.dumps(obj))` will yield a different object, assigned to different memory than the original `obj`. Thus, the typical round-trip test `obj == pickle.loads(pickle.dumps(obj))` fails, unless you just check that the `obj.__dict__` matches. However, that breaks some things like... [this line](https://github.com/pydata/patsy/blob/master/patsy/eval.py#L415)

I've also run across some issues with the stateful transforms like `center`, `standardize`, etc. These are functions (and from my rudimentary knowledge of computer science, pickling functions is generally a no-no). I need to set the `__qualname__` and `__name__` of these functions because when you generally do... 

```python
> center = stateful_transform(Center)
> center.__qualname__
'Center'
```

But when you `pickle.dump`, it realizes that `Center` refers to the class, and `center` is a function. So I have to hard-change the `__qualname__` to get this to work. [See here](https://github.com/pydata/patsy/blob/master/patsy/eval.py#L415).  This seems undesireable.  

Anyways, let me know if there's any interest in this work. I'm by no means an expert at this... and the test cases could obviously be improved. But I'm not a programmer by trade, so please grant me some forbearance. 

Thanks!


thequackdaddy:
Oh... and if you care to see the test-suite results... I ran it here...
https://travis-ci.org/thequackdaddy/patsy/jobs/219101866

Some small things fail here, but I think it works overall. 

Sorry, no coverage available.

patsy/categorical.py
```diff
@@ -46,20 +46,37 @@
...
-    __getstate__ = no_pickling
+    def __getstate__(self):
+        data = getattr(self, 'data')
+        contrast = getattr(self, 'contrast')
+        levels = getattr(self, 'levels')
+        return (0, data, contrast, levels)
+
+    def __setstate__(self, pickle):
+        version, data, contrast, levels = pickle
+        check_pickle_version(version, 0, name=self.__class__.__name__)
```
njsmith: You have to check the pickle version before unpacking it – e.g. if we ever added a 5th field to the pickle state here, then this would error out on the unpacking line, before getting to the `check_pickle_version` line. Basically we have to commit to never change anything that happens before we look at the version, so the first line should be something like
```python
check_pickle_version(pickle[0], 0, ...)   # if we commit to using a (version, ...) tuple forever
```
or
```python
check_pickle_version(pickle["version"], 0, ...)
```
if we decide maybe we want to be stuck with something more readable than a tuple ;-)

patsy/desc.py
```diff
@@ -166,7 +186,10 @@ def from_formula(cls, tree_or_string):
         assert isinstance(value, cls)
         return value
 
-    __getstate__ = no_pickling
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
+    # __getstate__ = no_pickling
```
njsmith: So this class is using the default `__getstate__`...?

patsy/desc.py
```diff
@@ -209,7 +234,7 @@ def _pretty_repr_(self, p, cycle): # pragma: no cover
                                 [self.intercept, self.intercept_origin,
                                  self.intercept_removed, self.terms])
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Why does `IntermediateExpr` need to support pickling? I think it should only exist as a temporary object in memory?

patsy/design_info.py
```diff
@@ -120,7 +121,17 @@ def __repr__(self):
             kwlist.append(("categories", self.categories))
         repr_pretty_impl(p, self, [], kwlist)
 
-    __getstate__ = no_pickling
```
njsmith: This class is also using the default `__getstate__`?

patsy/design_info.py
```diff
@@ -234,7 +245,10 @@ def _repr_pretty_(self, p, cycle):
                           ("contrast_matrices", self.contrast_matrices),
                           ("num_columns", self.num_columns)])
 
-    __getstate__ = no_pickling
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
+    # __getstate__ = no_pickling
```
njsmith: Likewise...

patsy/eval.py
```diff
@@ -10,7 +10,7 @@
 # for __future__ flags!
 
 # These are made available in the patsy.* namespace
-__all__ = ["EvalEnvironment", "EvalFactor"]
+__all__ = ["EvalEnvironment", "EvalFactor", "VarLookupDict"]
```
njsmith: Why does this add `VarLookupDict` to patsy's public API?

patsy/infix_parser.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, print_as):
     def __repr__(self):
         return "%s(%r)" % (self.__class__.__name__, self._print_as)
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Why does this need to support pickling?

patsy/infix_parser.py
```diff
@@ -70,7 +70,7 @@ def _repr_pretty_(self, p, cycle):
             kwargs = [("extra", self.extra)]
         return repr_pretty_impl(p, self, [self.type, self.origin], kwargs)
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Why does this need to support pickling?

patsy/infix_parser.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, type, token, args, origin):
     def _repr_pretty_(self, p, cycle):
         return repr_pretty_impl(p, self, [self.type, self.token, self.args])
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Likewise...

patsy/infix_parser.py
```diff
@@ -95,14 +95,14 @@ def __repr__(self):
         return "%s(%r, %r, %r)" % (self.__class__.__name__,
                                    self.token_type, self.arity, self.precedence)
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Likewise...

patsy/infix_parser.py
```diff
@@ -95,14 +95,14 @@ def __repr__(self):
...
 
 class _StackOperator(object):
     def __init__(self, op, token):
         self.op = op
         self.token = token
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Likewise...

patsy/infix_parser.py
```diff
@@ -115,7 +115,7 @@ def __init__(self, unary_ops, binary_ops, atomic_types, trace):
         self.atomic_types = atomic_types
         self.trace = trace
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Likewise...

patsy/origin.py
```diff
@@ -112,9 +112,11 @@ def __repr__(self):
             self.code[self.end:],
             self.start, self.end)
 
+    """
     # We reimplement patsy.util.no_pickling, to avoid circular import issues
     def __getstate__(self):
         raise NotImplementedError
+    """
```
njsmith: This needs an actual `__getstate__`

patsy/redundancy.py
```diff
@@ -73,7 +73,7 @@ def __repr__(self):
             suffix = "-"
         return "%r%s" % (self.factor, suffix)
 
-    __getstate__ = no_pickling
+    # __getstate__ = no_pickling
```
njsmith: Why does this need to support pickling?

patsy/state.py
```diff
@@ -76,7 +76,18 @@ def stateful_transform_wrapper(*args, **kwargs):
 # class QuantileEstimatingTransform(NonIncrementalStatefulTransform):
 #     def memorize_all(self, input_data, *args, **kwargs):
 
-class Center(object):
+
+class StatefulTransform(object):
+    def __getstate__(self):
+        return (0, self.__dict__)
```
njsmith: Umm. This is cheating :-)

patsy/test_pickling.py
```diff
@@ -0,0 +1,131 @@
...
+pickling_testcases = {
+    "evalfactor_simple": EvalFactor("a+b"),
+    "varlookupdict_simple": VarLookupDict([{"a": 1}, {"a": 2, "b": 3}]),
+    "evalenv_simple": EvalEnvironment([{"a": 1}]),
+    "evalenv_transform_center": EvalEnvironment([{'center': center}]),
+    "evalenv_transform_scale": EvalEnvironment([{'scale': scale}]),
+    "evalenv_transform_standardize": EvalEnvironment([{
+            'standardize': standardize
+    }]),
+    "evalenv_transform_catgorical": EvalEnvironment([{'C': C}]),
```
njsmith: missing "e" in "categorical"

patsy/test_pickling.py
```diff
@@ -0,0 +1,131 @@
...
+from patsy.splines import bs
+from patsy.desc import Term, ModelDesc
+from patsy.mgcv_cubic_splines import cc, te, cr
+from patsy.contrasts import ContrastMatrix
+from patsy.constraint import LinearConstraint
+from patsy.missing import NAAction
+from patsy.origin import Origin
+
+
+PICKE_TESTCASES_ROOTDIR = os.path.join(os.path.dirname(__file__), '..', 'pickle_testcases')
```
njsmith: missing "L" in "PICKLE"

patsy/util.py
```diff
@@ -732,3 +732,23 @@ def test_safe_string_eq():
...
+        error_msg = "This pickle is too old and not supported by this version of patsy anymore"
+    else:
+        return
+
+    if name:
+        error_msg += " (for object {})".format(name)
+    error_msg += "."
+
+    # TODO Use a better exception than ValueError.
+    raise ValueError(error_msg)
```
njsmith: I guess we could use `PickleError`

njsmith:
General comments:

* It's very annoying that pickle doesn't accept module objects :-(. The code for working around that seems basically reasonable, though maybe could be a bit cleaner.

* I don't understand why stateful transform functions like `center` need to be pickleable? the instantiated transform objects certainly do (and they need real `__getstate__` methods :-P), but a fitted model shouldn't be referring to the functions themselves anymore, I would think... maybe we need to refactor a bit to make this true.

* I'm not a big fan of all the `__eq__` changes – this is a globally visible change, and the definitions are just targeted to the idiosyncratic needs of our tests :-(. (And btw, they're all buggy, because in python you have to override `__ne__` too... I forget the exact rules, this might be fixed in py3, but definitely in py2.) Maybe a better approach would be to write end-to-end tests, that run some data through an unpickled model and checks that the result matches?

njsmith:
But overall: this isn't ready yet, but it's very exciting to see someone working on it :-)

thequackdaddy:
Thanks. I'll try and take a look at this tomorrow.

I think you caught a few of the shortcuts I was trying... I'll clean up as appropriate.

As the the __eq__ issue... i was also thinking I could define my own assert statement that would check that the __dict__ and all nested __dict__ were equal. Maybe call it "assert_pickle_equal". Some items, like EvalEnvironment, I don't see how to pass data through.

Thanks for your excellent comments. Hopefully I'll have an update tomorrow afternoon-ish.

thequackdaddy:
Update for today:

https://travis-ci.org/thequackdaddy/patsy/builds/219481856 (all green)

I've gone through/reverted your code as you recommended. Thanks for your pointers. 

I haven't figured out why stateful transform function like `center` are in the `EvalEnvironment`. I quickly tried to remove it, but got an error because patsy later looked for it. (Not sure if it actually uses it... need to follow the code some more.)

But this seems do-able.

thequackdaddy:
This has been very educational. Thanks for the help. 

I took a stab at fixing the issue where the patsy stateful transforms required `__qualname__` to be speicified. I think that once the `design_info` has been built, it should be safe to just remove the functions.... So I added a `clean` method on `EvalEnvironment` that removes them. 

[Here's](https://gist.github.com/thequackdaddy/159a805f735e8a0159349b906b7c5478) a gist showing that I think this works.

njsmith:
That looks like it will work, but I think it should be possible to solve in a more fundamental way. Looking at `memorize_passes_needed` in `EvalFactor` I think I see the problem: there's some code that pulls out the set of names referenced in the code (using the `ast_names` helper function), and then reduces the `EvalEnvironment` to just those names (`eval_env.subset`). This was added in #66 as preparation for the pickling support -- it's why trying to pickle an `EvalEnvironment` doesn't try to save your entire Python environment :-).

And then if you look a few lines *after* we compute the set of variables to keep, then there's some code that rewrites the code to convert stateful transform calls like `center(x)` into method calls on a newly allocate transform object.

So I think the best solution is to swap the order of these things: if we make the list of variables to keep *after* eliminating references to stateful transforms, then it should just work.

I know this is getting into somewhat deep wizardry... I'm going to try to look at this more closely within the next few days, and I can fix this part up if you don't get there first :-)

thequackdaddy:
You are correct... simply swapping out the order of when the subset is done (and doing the subset on the replaced funcall code) seems like it does the trick. 

There are 2 tests that would fail because of this change because they look for the stateful transforms in the `eval_env`. 

[here](https://github.com/pydata/patsy/pull/104/files#diff-f815d6e275a6eae8aa9e8d43d79a50c5L596) and [here](https://github.com/pydata/patsy/pull/104/files#diff-f815d6e275a6eae8aa9e8d43d79a50c5L652)

So I've changed the tests to do the opposite... assert that the keys aren't in the `namespaces`. 

[And here's travis](https://travis-ci.org/thequackdaddy/patsy/builds/219959868)

thequackdaddy:
@njsmith Wanted to check in and see if you had any other thoughts on this. Seems like it works for me but happy to work it more if you think this is something that is merge-able

thequackdaddy:
So I built this on a windows server 2016 machine and I got a couple of nosetest failures:

```
======================================================================
ERROR: patsy.test_pickling.test_pickling_old_versions_still_work('C:\\Users\\pquackenbush\\git\\patsy\\patsy\\..\\pickle_testcases\\0.5\\transform_center.pickle',)
----------------------------------------------------------------------
  File "C:\ProgramData\Anaconda3\envs\bleeding\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\pquackenbush\git\patsy\patsy\test_pickling.py", line 137, in check_pickling_old_versions_still_work
    pickle.load(f))
TypeError: data type "f16" not understood

======================================================================
ERROR: patsy.test_pickling.test_pickling_old_versions_still_work('C:\\Users\\pquackenbush\\git\\patsy\\patsy\\..\\pickle_testcases\\0.5\\transform_standardize_norescale.pickle',)
----------------------------------------------------------------------
  File "C:\ProgramData\Anaconda3\envs\bleeding\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\pquackenbush\git\patsy\patsy\test_pickling.py", line 137, in check_pickling_old_versions_still_work
    pickle.load(f))
TypeError: data type "f16" not understood

======================================================================
ERROR: patsy.test_pickling.test_pickling_old_versions_still_work('C:\\Users\\pquackenbush\\git\\patsy\\patsy\\..\\pickle_testcases\\0.5\\transform_standardize_rescale.pickle',)
----------------------------------------------------------------------
  File "C:\ProgramData\Anaconda3\envs\bleeding\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\pquackenbush\git\patsy\patsy\test_pickling.py", line 137, in check_pickling_old_versions_still_work
    pickle.load(f))
TypeError: data type "f16" not understood

----------------------------------------------------------------------
Ran 50 tests in 0.281s
```

Any hints before I start digging into this? 

I wonder if this is a hint I should use appveyor too...

thequackdaddy:
@njsmith Ok I dug into this a bit... 

I think the issue is that 'f16' is the dtype for `np.float128`. The computer I ran this on--at the time--had `np.float128` available, but my Windows Server 2016 computer does not have that dtype available. (I'm a little confused because I can't seem to figure out what I did that got the `float128` datatype to show up in the first place... Both are 64 bit systems... one is a laptop, one is a xeon e5 virtual machine). 

patsy tries to promote the data to a "wider" dtype [here](https://github.com/pydata/patsy/blob/master/patsy/util.py#L294). So if you make the pickle on a computer with `np.float128`, then try to load it on a computer without that available, I think this will happen. 

Options:
* Maybe only promote up to `np.float64`. 
* Have an error handle that says when you load a datatype that doesn't exist on the current environment, to gently fail. 
* re-write tests to not have this issue

Thoughts?

thequackdaddy:
@njsmith Pinging to see if there is any interest in reviewing/merging this. 

Thanks!

thequackdaddy:
@njsmith Hello again. Just pinging to see if there's interest in this... would be a little helpful for me to cut down on some model sizes.

thequackdaddy:
@njsmith - Checking in again to see if this is on the right track/something that could be merged. Let me know either way. Thanks so much!

dswah:
@thequackdaddy @njsmith 
how is this PR looking?

thequackdaddy:
I'm waiting for review/feedback... This is definitely a big project, and part of it--such as the whole `assert_pickled_equals` is difficult to check. 

If you want to read/through, provide feedback, I think that might be helpful.

matthewwardrop:
Hi @thequackdaddy ,

Thanks for these patches. As one of the new maintainers I'm going through the PRs and figuring out what still needs attention. Even though your patches have been sitting here for along time, the code-base hasn't change much and so it's likely they are still suitable. Is this something you are still interested in hashing out?

Given that the name of the game from a maintenance perspective is to keep things working as is, I'm a little cautious to introduce massive changes that might introduce new bugs. It should also be noted that https://github.com/matthewwardrop/formulaic (the spiritual successor of this project) already has support for pickling, as well as sparse representations and is much more performant. But if you'd like to explore finishing up this work, I'm willing to dig into it with you.

Best,
M

PR #131, opened 2018-10-01
fix compatibility with unreleased changes to stdlib tokenizer

mwhudson:
https://github.com/python/cpython/commit/c4ef4896eac86a6759901c8546e26de4695a1389
(not yet in any released version, but it's been backported to all
versions of Python in git, even 2.7!) changed the behaviour in the
stdlib's tokenize module to emit a synthetic NEWLINE token even if the
file does not end with a newline. This was causing a spurious
"mixed-line-endings" warning to be emitted, but luckily the synthetic
token is easy to test for (the token text is "").

### DIFF
--- a/patsy/tokens.py
+++ b/patsy/tokens.py
@@ -32,7 +32,10 @@ def python_tokenize(code):
             if pytype == tokenize.ENDMARKER:
                 break
             origin = Origin(code, start, end)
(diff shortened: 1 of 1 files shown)


njsmith:
Thanks! I should fix those tests, but it looks like the failures are just deprecation warnings, so I'm going t omerge anyway...

PR #133, opened 2018-10-14
MAINT: Travis fixes

thequackdaddy:
Here's a way to fix travis. 

There's a conda dependency problem that can be easily solved by building the conda environment with only one `conda install`. 

`numpy.testing.util` is deprecated as well, so I did a `try/except` to use the newer one, if available. 

One concern is that numpy is deprecating `np.matrix`. 

```
======================================================================
ERROR: patsy.util.test_atleast_2d_column_default
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/patsy/util.py", line 176, in test_atleast_2d_column_default
    assert type(atleast_2d_column_default(np.matrix(1))) == np.ndarray
  File "/home/travis/miniconda/envs/testenv/lib/python3.6/site-packages/numpy/matrixlib/defmatrix.py", line 118, in __new__
    PendingDeprecationWarning, stacklevel=2)
PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
```

It might be time to deprecate it here too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,10 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
-  - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION numpy scipy coverage nose pip
+  - export PKGS="numpy scipy coverage nose pip"
+  - if [ "$PANDAS_VERSION_STR" != "NONE" ]; then export PKGS="${PKGS} pandas${PANDAS_VERSION_STR}"; fi
+  - conda create -q -n testenv python=$TRAVIS_PYTHON_VERSION ${PKGS}
   - source activate testenv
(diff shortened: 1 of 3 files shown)


njsmith:
This is awesome, thanks!

The only way we use `matrix` is to test that we don't blow up if a user is unwisely using `matrix` and passes it into patsy, so I think keep tests for that is reasonable, and anyone doing this is already getting deprecations from numpy.

It would definitely make sense to drop support for ancient pandas versions, and would also simplify the testing... we could probably even drop support for the numpy versions that don't have `numpy.testing.assert_equals`.

Right now I just want to merge this to get our CI working again so I can make a release to get #131 out... let's see if I remember how to do a release!

PR #140, opened 2018-10-29
TST: Add python 3.7 testing

thequackdaddy:
I think this will get python 3.7 testing on patsy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: none
+language: minimal
 matrix:
   include:
     - env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,16 @@ matrix:
       env: PANDAS_VERSION_STR="NONE"
     - python: 2.7
       env: PANDAS_VERSION_STR="NONE"
+    - python: 3.7
+      sudo: true
+      dist: xenial
+    - python: 3.7
+      sudo: true
+      dist: xenial
+      env: PANDAS_VERSION_STR="NONE"
 
 # This disables sudo, but makes builds start much faster
 # See http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
-sudo: false
 before_install:
   # Work around terrible pathological behaviour in OpenBLAS multithreading, that causes execution time to blow up from 3 minutes to 18 minutes, apparently in SVD on smallish matrices
   - export OMP_NUM_THREADS=1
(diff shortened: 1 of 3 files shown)


thequackdaddy:
I also added a `try ... except` around the `import collections.abc from Mapping` as that got moved.

tools/check-API-refs.py
```diff
@@ -9,7 +9,7 @@
 root = dirname(dirname(abspath(__file__)))
 patsy_ref = root + "/doc/API-reference.rst"
 
-doc_re = re.compile("^\.\. (.*):: ([^\(]*)")
```
njsmith: Better to make it a raw string, `r"..."`

.travis.yml
```diff
@@ -23,6 +23,13 @@ matrix:
       env: PANDAS_VERSION_STR="NONE"
     - python: 2.7
       env: PANDAS_VERSION_STR="NONE"
+    - python: 3.7
+      sudo: true
+      dist: xenial
+    - python: 3.7
+      sudo: true
+      dist: xenial
+      env: PANDAS_VERSION_STR="NONE"
```
njsmith: This is the simplest change that will work, but the overall situation is kind of silly. The only reason `sudo: true, dist: xenial` is needed is to get travis's version of 3.7, but we don't use travis's version of 3.7 for anything at all. We request it, force Travis to install it, and then we immediately throw it away and switch to conda. (And this is true for all the Python versions, just it becomes more obviously silly at the point where we start adding workarounds for limitations of Travis's Python...)

We'd be better off using `language: none`, and setting `PYTHON_VERSION` as an env-var... that's a larger change though.

tools/check-API-refs.py
thequackdaddy: Changed.

.travis.yml
thequackdaddy: Agreed. I went ahead and pushed how I think you could do it without the hack. Honestly, its not much work and not that much of a larger change, so i'd suggest using it. You'll likely want to squash all my commits.

njsmith:
Apparently I misremembered the syntax, and `language: none` actually means "ruby". The actual no-language images are `minimal` and `generic`: https://docs.travis-ci.com/user/languages/minimal-and-generic/

I just tried switching it to `minimal`, let's see what Travis does...

Juanlu001:
Any chance to release these changes?

PR #153, opened 2020-01-15
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,27 +38,98 @@ matrix:
       - PYTHON_VERSION=3.7
       - PANDAS_VERSION_STR="NONE"
 
+    # 0.14.0 is the last version with the old categorical system
+    # libfortran=1.0 is needed to work around a bug in anaconda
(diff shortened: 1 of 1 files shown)


bashtage:
Why do you think an ARM build is needed? Patsy is pure Python.

rhenwood-arm:
@bashtage : I have been working with @ossdev07 on a down-stream project on AArch64 (ARM64) that has patsy as a dependency. As you point out, patsy is a pure python project with a straightforward Travis-ci build job and that doesn't have a binary release. In this case, there is very little benefit in testing on alternative architectures, and this PR can probably be held in reserve until it is needed.

matthewwardrop:
Thanks for the patches! As a pure python library, we don't need to test against specific architectures unless we need to work around bugs in upstream (C-level) packages. I'll close this one for now, but feel free to reopen and rebase if it becomes necessary.

PR #162, opened 2020-09-08
do not depend on nose

pgajdos:
Use pytest's raise instead of nose's assert_raise. nose is deprecated, will not work with future python versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ script:
   # The --exe is because python sometimes marks all installed modules
   # as executable, so without --exe nosetests will just ignore
   # everything. Baffling, but so it goes.
-  - coverage run --source=$INSTALLDIR --rcfile=../.coveragerc $(which nosetests) -vvv --nocapture --exe --failure-detail --all-modules $INSTALLDIR
+  - coverage run --source=$INSTALLDIR --rcfile=../.coveragerc $(which pytest) --cov patsy $INSTALLDIR
   - coverage report --rcfile=../.coveragerc --show-missing
   - python ../tools/check-API-refs.py
 notifications:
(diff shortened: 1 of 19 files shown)


bashtage:
While I support this patch, you might want to see if formulaic might suit your needs.  I think patsy is mostly in EOL mode.

pgajdos:
It seems that few packages in openSUSE still require python-patsy:
   python-ggplot
   python-seaborn
   python-statsmodels

bashtage:
seaborn needs it because of statsmodels IIRC. statsmodels is considering whether formulaic can become the formula parser.

ggplot is abandoned without a commit in 4 years. plotnine is the preferred grammar for graphics in Python.

bashtage:
An alternative fix (and more standard) would be to pull all of the tests out of code files into test files so that nose isn't needed to run patsy.

pgajdos:
> seaborn needs it because of statsmodels IIRC. statsmodels is considering whether formulaic can become the formula parser.
> 
> ggplot is abandoned without a commit in 4 years. plotnine is the preferred grammar for graphics in Python.

Thanks for letting me know. I just go trough python modules that still use nose and try to fix them, without not much knowledge about the context.

pgajdos:
> An alternative fix (and more standard) would be to pull all of the tests out of code files into test files so that nose isn't needed to run patsy.

@bashtage, I am not sure I understand. Of course it would be better to have them tests in separate test files, but you use assert_raises, so you will have to youse nose anyway, no?

Anyway, it seems that travis does not like me today, so let's try later.

pgajdos:
Closing for now.

bashtage:
Normally projects have directories that look like

root\patsy
root\patsy\tests
root\setup.py

and then in root\patsy\tests you have files that are named `test_<something meaningful>.py`.  Under normal execution files under root\patsy\test are not executed, and so are only picked up by the test runner, e.g., nose.  This means that patsy will be importable under Python 3.9 even if nose it not. It won't be testable though.

PR #163, opened 2020-09-09
use pytest instead of nose

pgajdos:
nose is not maintained and will stop working in future python versions. Use pytest as a replacement.

The build will fail, will try to sort out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ before_install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
-  - export PKGS="numpy scipy coverage nose pip"
+  - export PKGS="numpy scipy coverage nose pip pytest"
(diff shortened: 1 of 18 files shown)


pgajdos:
Builds that now fail require not available versions of python and/or pandas.

I do not feel brave enough to switch from nosetests to pytest call in .travis.yml coverage commandline, but it is not necessary until you do not build against newer pythons (especially when you feel patsy EOL as per https://github.com/pydata/patsy/pull/162#issuecomment-688831172).

pgajdos:
Note that I run these tests with pytest (without nose available) in %check section without any issues. For coverage, probably something like pytest-cov would have to be installed.

bashtage:
You should remove nose as a dependency and only use pytest to run the tests.

matthewwardrop:
Closed by #178. Thanks so much for your patches.

PR #164, opened 2020-09-09
TST: Use pytest as the test runner

bashtage:
Switch to pytest
Remove testing of legacy Python versions < 3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,13 +49,16 @@ install:
 script:
   # We change directories to make sure that python won't find the copy
   # of patsy in the source directory.
-  - mkdir empty
-  - cd empty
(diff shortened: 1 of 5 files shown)


bashtage:
xref #163

bashtage:
@njsmith Do you have any thoughts on trying to move to pytest?  This PR runs the same tests with more-or-less the same coverage.

matthewwardrop:
Hi @bashtage and @pgajdos ; moving to `pytest` makes eminent sense; and since we are more or less in a feature freeze now removing legacy versions makes sense. I'll take a closer look at this PR and #163 shortly. Neither drops `nose` entirely, and have somewhat complementary changes. I will likely end up synthesizing these changes together with my own, and give the appropriate credit.

Thanks for these patches!

matthewwardrop:
Closed by #178. Thanks so much for your patches.

PR #169, opened 2021-05-04
cast df output to np.int8 to save memory

JTLehman1:


JTLehman1:
I am not trying to change pydata:master Apparently I do not understand how forks work.  My goal is only to make a change that I can pip install for my own use case.

PR #173, opened 2021-07-30
License file to be included into metadata dir

milos-korenciak:
Hi,
Please, add the LICENSE.txt file into metadata dir of your package (e.g. by accepting this pull request).
I understand you probably hate software laws etc., so I will be short.

You are using BSD 2-Clause license. It says also: "Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer." But your whole package distributed on PyPI.org does not contain even one. This short change will add the file LICENSE.txt (with text of BSD license) into metadata of your package (= into e.g. .../site-packages/patsy-0.5.1.dist-info/LICENSE.txt of your virtualenv). This way you are 100% clear and comply with your chosen license.

Thank you for your time and your work.
Pax et bonum.


matthewwardrop:
Thanks for your patch! This does not seem to be an issue with modern versions of setuptools (this is done automatically now: https://github.com/pypa/setuptools/pull/1536); and I have confirmed that everything is working. Nevertheless, I see no harm in explicitly calling this out in the metadata, so I'll merge in this patch in anyway.

PR #187, opened 2022-08-02
Handle dropped future features

### LINKED ISSUE #186: Importing `patsy.eval` fails on Python 3.11
Because the `annotations` feature was postponed indefinitely, `__future__.annotations.getMandatoryRelease()` returns `None` in Python 3.11, which [means a planned feature got dropped](https://docs.python.org/3.11/library/__future__.html).

As a result, importing `patsy.eval` fails with a traceback like:

```
  File "<string>", line 1, in <module>
  File "/home/ben/src/forks/patsy/patsy/__init__.py", line 77, in <module>
    import patsy.highlevel
  File "/home/ben/src/forks/patsy/patsy/highlevel.py", line 19, in <module>
    from patsy.design_info import DesignMatrix, DesignInfo
  File "/home/ben/src/forks/patsy/patsy/design_info.py", line 38, in <module>
    from patsy.desc import ModelDesc, Term
  File "/home/ben/src/forks/patsy/patsy/desc.py", line 14, in <module>
    from patsy.eval import EvalEnvironment, EvalFactor
  File "/home/ben/src/forks/patsy/patsy/eval.py", line 36, in <module>
    _ALL_FUTURE_FLAGS = _all_future_flags()
                        ^^^^^^^^^^^^^^^^^^^
  File "/home/ben/src/forks/patsy/patsy/eval.py", line 32, in _all_future_flags
    if feature.getMandatoryRelease() > sys.version_info:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'NoneType' and 'sys.version_info'
```

PR to follow.

musicinmybrain:
Fixes #186 by handling future features where the mandatory release version is `None`, indicating the feature was dropped. These are now included in `patsy.eval._ALL_FUTURE_FLAGS` rather than causing a `TypeError`.

The first appearance of such a feature is in Python 3.11, in which `__future__.annotations.getMandatoryRelease()` returns `None`.

### DIFF
--- a/patsy/eval.py
+++ b/patsy/eval.py
@@ -29,7 +29,11 @@ def _all_future_flags():
     flags = 0
     for feature_name in __future__.all_feature_names:
         feature = getattr(__future__, feature_name)
-        if feature.getMandatoryRelease() > sys.version_info:
+        mr = feature.getMandatoryRelease()
+        # None means a planned feature was dropped, or at least postponed
+        # without a final decision; see, for example,
+        # https://docs.python.org/3.11/library/__future__.html#id2.
+        if mr is None or mr > sys.version_info:
             flags |= feature.compiler_flag
     return flags
 


patsy/eval.py
```diff
@@ -29,7 +29,11 @@ def _all_future_flags():
     flags = 0
     for feature_name in __future__.all_feature_names:
         feature = getattr(__future__, feature_name)
-        if feature.getMandatoryRelease() > sys.version_info:
+        mr = feature.getMandatoryRelease()
+        # None means a planned feature was dropped, or at least postponed
+        # without a final decision; see, for example,
+        # https://docs.python.org/3.11/library/__future__.html#id2.
+        if mr is None or mr > sys.version_info:
```
bashtage, line 36: If is dropped, should the next line run?

musicinmybrain: Answering that requires some reflection on what this code is trying to accomplish.

In this case, the PEP corresponding to the `annotations` feature was dropped in the sense that it was postponed twice, and a decision has not been made about whether it will be scheduled for a future release or rejected altogether. However, the feature is still available in `__future__`, and it might be scheduled for stabilization in the future.

I figured that this code is looking for future features that *could have been imported* and *are not included in the current Python release*. It seems like features with mandatory release `None` still satisfy this condition. It’s possible that I’m not seeing the full context, though.

stonebig: the patch maked patsy work on my python-3.11.0rc1 test

matthewwardrop: From what I can see it looks like this code is trying to capture all possible flags that could be set by the user; so I think keeping the feature when `mr is None` makes sense.

matthewwardrop (approved):
Thanks for the patch @musicinmybrain! I'll merge this in and extend coverage to the Python 3.11 release, fixing any other issues that come up; and then put out a new patch release.

musicinmybrain:
Thanks!

EwoutH:
Thanks for fixing this!

Would it be possible to get a new (patch) release with this fix included? Currently over at [statsmodels](https://github.com/statsmodels/statsmodels) our [tests are breaking](https://dev.azure.com/statsmodels/statsmodels-testing/_build/results?buildId=4662&view=logs&j=e5ec9234-c220-5b28-4136-a9c82efad0a5&t=314f1936-9e6b-5410-c4ea-c53498aa77ae&l=63) for Python 3.11 due to this issue.

matthewwardrop:
Hi @EwoutH , Thanks for the nudge! With scipy having started supporting 3.11, we could now push out support also. I've pushed out the 0.5.3 release.

Note: It would be nice to migrate `statsmodels` over to [Formulic](https://github.com/matthewwardrop/formulaic) at some point soon :). Generally it is more robust, has more features, and is much faster.

EwoutH:
> Hi @EwoutH , Thanks for the nudge! With scipy having started supporting 3.11, we could now push out support also. I've pushed out the 0.5.3 release.

Thanks a lot for the quick response and the new release! I'm happy to report it unbreaks our Python 3.11 CI for Ubuntu and macOS! (Windows still has a separate issue)

> Note: It would be nice to migrate `statsmodels` over to [Formulaic](https://github.com/matthewwardrop/formulaic) at some point soon :). Generally it is more robust, has more features, and is much faster.

It sounds great! I'm not a core maintainer of statsmodels however, should we open an issue there to discuss it with them?

matthewwardrop:
> It sounds great! I'm not a core maintainer of statsmodels however, should we open an issue there to discuss it with them?

Hmm... I guess I should publish v1.0.0 of formulaic first (which will come with more complete documentation). In the mean time, patsy will continue working :). Thanks @EwoutH !

PR #192, opened 2023-04-26
pass whole module code over 2to3 filter

kloczek:
Remove python 2.x syntax by use 2to3 script.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,27 +1,27 @@
 # -*- coding: utf-8 -*-
 
 # General information about the project.
-project = u'patsy'
-copyright = u'2011-2015, Nathaniel J. Smith'
+project = 'patsy'
+copyright = '2011-2015, Nathaniel J. Smith'
 
 import sys
-print "python exec:", sys.executable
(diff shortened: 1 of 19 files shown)


bashtage:
A few substitutions can be skipped it looks like it.  The unicode to str all appear unnecessary, as do the long to int.  Also many of the print statement changes are inserting unnecessary parentheses.

kloczek:
OK NP .. so just please close this PR and pass the code over 2to3 filter and fix manually what you thing that should look a biot different 😋

bashtage:
You can just make some changes and commit them in a new commit.  patsy is in long term maintenance mode, and will eventually fall over in pandas 2.x when `is_categorical_dtype` is removed, unless that is patched.   The best odds of getting a PR accepted would be to make it as minimal as possible.

kloczek:
Issue is that use python 2.x syntax blocks two things:
- produce bytecode using latest python 3.x
- build documentation using latest sphinx 6.x

Sphinx fails even with that PR but to fix that issue as well this one needs to be sorted out first.

Issue #196, opened 2023-11-01
Test failures with Python 3.12


mgorny:
Tested on 0.5.3 and 6bd31ca7f5f12e77a918681d14b7c28d931bde38.

```pytb
$ tox -e py312
py312: install_deps> python -I -m pip install '.[test]'
.pkg: install_requires> python -I -m pip install 'setuptools>=40.8.0' wheel
.pkg: _optional_hooks> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_wheel> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: install_requires_for_build_wheel> python -I -m pip install wheel
.pkg: prepare_metadata_for_build_wheel> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
py312: install_package_deps> python -I -m pip install 'numpy>=1.4' six
py312: install_package> python -I -m pip install --force-reinstall --no-deps /tmp/patsy/.tox/.tmp/package/1/patsy-0.5.3+dev.tar.gz
py312: commands[0]> pytest -vv --cov=patsy --cov-config=/tmp/patsy/.coveragerc --cov-report=term-missing --cov-report=xml --cov-report=html:/tmp/patsy/.tox/coverage/py312
========================================================= test session starts =========================================================
platform linux -- Python 3.12.0, pytest-7.4.3, pluggy-1.3.0 -- /tmp/patsy/.tox/py312/bin/python
cachedir: .tox/py312/.pytest_cache
rootdir: /tmp/patsy
configfile: setup.cfg
testpaths: patsy
plugins: cov-4.1.0
collected 148 items                                                                                                                   

patsy/build.py::test__max_allowed_dim PASSED                                                                                    [  0%]
patsy/build.py::test__eval_factor_numerical PASSED                                                                              [  1%]
patsy/build.py::test__eval_factor_categorical PASSED                                                                            [  2%]
patsy/build.py::test__column_combinations PASSED                                                                                [  2%]
[...]

PR #197, opened 2023-11-30
Add support for Python 3.12 (and adjust code for tokenizer changes).

matthewwardrop:
closes: #196


bashtage:
Is there any expectations the CI runs or needs to be passed?

matthewwardrop:
> Is there any expectations the CI runs or needs to be passed?

Yes... but it is struggling because Python 2.7 and 3.6 no longer exist in GitHub actions support versions. Fixing.

PR #202, opened 2023-12-31
MAINT: Repalce row_stack with vstack

bashtage:
row_stack is official deprecated in NumPy 2, and will be dropped eventually.  Currently produces a warning


matthewwardrop:
LGTM. Thanks @bashtage .

PR #199, opened 2023-12-18
remove pandas deprecation warning

EpigeneMax:
Closes #198

Issue #198, opened 2023-12-18
Deprecation Warning from pandas


EpigeneMax:
`patsy` uses functions from pandas that are deprecated:

```
patsy/util.py:672: FutureWarning: is_categorical_dtype is deprecated and will be removed in a future version. Use isinstance(dtype, CategoricalDtype) instead
  return _pandas_is_categorical_dtype(dt)
```

I'll take a look and try to make a PR for this.

lesteve:
It does not look like this issue has gone completely gone away somehow. We see something similar with patsy 0.5.5 inside the scikit-learn CI where we turn warnings into errors for our gallery examples.

I have to say I don't fully follow the logic of #199, maybe I post more comments in the PR about my questions ... in our case it made it slightly worse since the warning is at `patsy` import time rather than only when you use `patsy.util.is_pandas_categorical` and it seems like we were only using basic seaborn functionality which did not need `patsy.util.is_pandas_categorical`.

To reproduce (pandas >= 2.1.0 which is the version that deprecated `is_categorical_dtype` see [this](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v2.1.0.html#other-deprecations)):
```
PYTHONWARNINGS=error::DeprecationWarning \
    python -c 'import patsy.util; patsy.util.safe_is_pandas_categorical_dtype(int)'
```

With patsy 0.5.5 you get the error during `patsy` import, with patsy 0.5.4 you get the warning when calling the `patsy.util.safe_is_pandas_categorical`.

lesteve:
One thing I noticed is that pandas may raise a `DeprecationWarning` or a `FutureWarning` depending on the pandas version. Apparently it was a `FutureWarning` from 2.1.0 until 2.1.3 and is now a `DeprecationWarning` in 2.1.4.

Context: https://github.com/pandas-dev/pandas/issues/54970 and https://github.com/pandas-dev/pandas/pull/56377.

EpigeneMax:
Thanks for reporting this, @lesteve.

The patch in #199 indeed moves the warning at import time, and only catches `FutureWarning`, not `DeprecationWarning`. With #203 I see no warning with `pandas` 2.1.4.

matthewwardrop:
I noted that as well, but assumed your patch was about making things work when this functionality is eventually removed, and less about suppressing the warnings themselves.... but I can totally see why you'd want to remove those warnings.

PR #203, opened 2024-01-03
properly handle pandas 2.1.4 deprecation warning

EpigeneMax:
Following additional comments in #198, this PR improves on #199.

### DIFF
--- a/patsy/util.py
+++ b/patsy/util.py
@@ -53,7 +53,7 @@ else:
             import warnings
             with warnings.catch_warnings(record=True) as w:
                 _pandas_is_categorical_dtype(int)
-                if len(w) > 0 and issubclass(w[-1].category, FutureWarning):
+                if len(w) > 0 and issubclass(w[-1].category, (DeprecationWarning, FutureWarning)):
                     _pandas_is_categorical_dtype = None
         else:
             _pandas_is_categorical_dtype = lambda x: isinstance(x, pandas.CategoricalDType)


lesteve:
Thanks for the PR, I'll try to check whether that gets rid of the warnings. In the mean time I have a few comments.

It would seem more straightforward to first try `isinstance(..., pd.CategoricalDtype)` since this is the recommended way since pandas 2.1.0 and I am guessing that it works for older pandas versions as well. Having said that I did not manage to find the relevant info about which pandas version supports it.

This does seem cleaner than trying to catch a warning with a deprecated code path and the warnings catching logic seems a bit brittle. Why catching the last warning rather than any warnings if there are more than one, although this is probably edge-casy.

Also when using the warnings catching logic, when `pd.types.api.is_categorical_dtype` gets removed from pandas you once again need to fix in `patsy` which does not seem great.

To finish with I don't think the code logic is correct. When you catch the warning, you set                     `_pandas_is_categorical_dtype = None`. I believe it was correct (although a bit convoluted) in your original PR #199 because you added a `if _pandas_is_categorical_dtype is None` afterwards, but a subsequent commit https://github.com/pydata/patsy/pull/199/commits/01c6b70fed8b6f00b85c79dedece01a297ad6f35 by @matthewwardrop broke it by removing the `if` clause.

All in all I do think that in 0.5.5 `patsy.util.safe_is_categorical_pandas` is currently broken with pandas 2.1.0 to 2.1.3:
```py
import pandas as pd
import patsy
print(f"{pd.__version__=}")
import patsy
print(f"{patsy.__version__=}")

categorical_dtype = pd.CategoricalDtype(['a', 'b'])
print(patsy.util.safe_is_pandas_categorical_dtype(categorical_dtype))
```

Output:
```
pd.__version__='2.1.3'
patsy.__version__='0.5.5'
  File "/tmp/test.py", line 8, in <module>
    print(patsy.util.safe_is_pandas_categorical_dtype(categorical_dtype))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lesteve/micromamba/lib/python3.11/site-packages/patsy/util.py", line 681, in safe_is_pandas_categorical_dtype
    return _pandas_is_categorical_dtype(dt)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not callable
```

matthewwardrop:
> It would seem more straightforward to first try isinstance(..., pd.CategoricalDtype) since this is the recommended way since pandas 2.1.0 and I am guessing that it works for older pandas versions as well. Having said that I did not manage to find the relevant info about which pandas version supports it.

Agreed. I took a quick look into this and didn't find it either when reviewing the original PR. I figured that the code would unbreak things in the indefinite future, and so wasn't too bothered.

> Also when using the warnings catching logic, when pd.types.api.is_categorical_dtype gets removed from pandas you once again need to fix in patsy which does not seem great.

+1.

> To finish with I don't think the code logic is correct. When you catch the warning, you set _pandas_is_categorical_dtype = None. I believe it was correct (although a bit convoluted) in your original PR https://github.com/pydata/patsy/pull/199 because you added a if _pandas_is_categorical_dtype is None afterwards, but a subsequent commit https://github.com/pydata/patsy/commit/01c6b70fed8b6f00b85c79dedece01a297ad6f35 by @matthewwardrop broke it by removing the if clause.

Guilty as charged. Should have looked over these changes a little bit more closely.

lesteve:
Sounds great, let me know if I can help!

EpigeneMax:
> Also when using the warnings catching logic, when pd.types.api.is_categorical_dtype gets removed from pandas you once again need to fix in patsy which does not seem great.

Guilty as charged, I did focus too much on detecting deprecation. I agree with you that the warning catching logic is not great, but I think it is good enough for this use case.

> It would seem more straightforward to first try isinstance(..., pd.CategoricalDtype) since this is the recommended way since pandas 2.1.0

Agreed, but I did not want to break the behavior on older versions of `pandas`.

I'll let @matthewwardrop go with a PR, let me know if I can help!

matthewwardrop:
Fixed in #204

lesteve:
Nice, thanks a lot for this and the 0.5.6 release!

Also, since there there is an error for patsy 0.5.5 and pandas>=2.1.0,<=2.1.3 as noted in https://github.com/pydata/patsy/pull/203#issuecomment-1874942671, it may be worth considering yanking patsy 0.5.5 from PyPI and marking the package as broken in conda-forge?

It could well be the case that this is not worth it, since that not that many people will bump into this issue and if they do, they will find this issue where the easy fix is to update patsy to 0.5.6.

matthewwardrop:
I don't think there was a whole tonne of utility, but I went ahead and yanked it from PyPI. Marking packages as broken on conda (or updating the metadata to require pandas < 2.1) is a little more involved, and not worth the effort, IMHO. If you feel differently, feel free to propose the changes via the appropriate conda-forge channels.

lesteve:
Thanks :pray:, this seems perfectly good enough.

I was not too sure it was worth it either, so I hope I did not twist your arm too much :wink:.

PR #204, opened 2024-01-04
Avoid deprecated pandas `is_categorical_dtype` API

matthewwardrop:
This patch side-steps the issues addressed by #198 and #199, but avoiding recently deprecated APIs.

### DIFF
--- a/patsy/util.py
+++ b/patsy/util.py
@@ -41,28 +41,16 @@ else:
 # pandas.
 have_pandas_categorical = (have_pandas and hasattr(pandas, "Categorical"))
 if not have_pandas:
-    have_pandas_categorical_dtype = False
     _pandas_is_categorical_dtype = None
 else:
-    if hasattr(pandas, "api"):
-        # This is available starting in pandas v0.19.0
-        have_pandas_categorical_dtype = True
+    if hasattr(pandas, "CategoricalDtype"):  # pandas >= 0.25
+        _pandas_is_categorical_dtype = lambda x: isinstance(getattr(x, "dtype", x), pandas.CategoricalDtype)
+    elif hasattr(pandas, "api"):  # pandas >= 0.19
         _pandas_is_categorical_dtype = getattr(pandas.api.types, "is_categorical_dtype", None)
-        # pandas.api.types._pandas_is_categorical_dtype is deprecated in pandas v2.1.0
-        if _pandas_is_categorical_dtype is not None:
-            import warnings
-            with warnings.catch_warnings(record=True) as w:
-                _pandas_is_categorical_dtype(int)
-                if len(w) > 0 and issubclass(w[-1].category, FutureWarning):
-                    _pandas_is_categorical_dtype = None
-        else:
-            _pandas_is_categorical_dtype = lambda x: isinstance(x, pandas.CategoricalDType)
-    else:
-        # This is needed for pandas v0.18.0 and earlier
+    else:  # pandas <=0.18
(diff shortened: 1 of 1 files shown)


patsy/util.py
```diff
@@ -45,18 +45,13 @@
...
-        # pandas.api.types._pandas_is_categorical_dtype is deprecated in pandas v2.1.0
-        if _pandas_is_categorical_dtype is not None:
-            import warnings
-            with warnings.catch_warnings(record=True) as w:
-                _pandas_is_categorical_dtype(int)
-                if len(w) > 0 and issubclass(w[-1].category, FutureWarning):
-                    _pandas_is_categorical_dtype = None
-        else:
-            _pandas_is_categorical_dtype = lambda x: isinstance(x, pandas.CategoricalDType)
+        _pandas_is_categorical_dtype = lambda x: isinstance(x, pandas.CategoricalDType)
```
bashtage, line 54: Pretty sure this will not work on old pandas, e.g., those that work with python 2.7

patsy/util.py
```diff
@@ -45,18 +45,13 @@
     _pandas_is_categorical_dtype = None
 else:
     if hasattr(pandas, "api"):
```
bashtage, line 47: I think you could start with `hasattr(pandas, "CategoricalDType")` and then the lambda

Failing this, you could do `elif hasattr(pandas, "api")`, and then `_pandas_is_categorical_dtype = getattr(pandas.api.types, "is_categorical_dtype", None)`, 

finally, else, `None`.

Something like 
```python
if hasattr(pandas, "CategoricalDType"):
    _pandas_is_categorical_dtype = lambda x: isinstance(x, pandas.CategoricalDType)
elif hasattr(pandas, "api"):
    if getattr(pandas.api.types, "is_categorical_dtype", None) is not None:
        def wrapped_is_categorigal_type(typ):
            with warnings.catch_warnings():
                warnings.filterwarnings('ignore')
                return pandas.api.types.is_categorical_dtype(typ)

        _pandas_is_categorical_dtype = wrapped_is_categorigal_type
else: 
    _pandas_is_categorical_dtype  = None
```

matthewwardrop: Yeah... still fiddling with this a bit... delving into pandas ancient history and seeing where things were introduced.

matthewwardrop: Why do we need the warnings filter?

bashtage: I reckon just throw it in to make sure noting comes up.  We know if is there then it will either FutureWarn, DepreactionWarn, or nothing. The first two are just noise.

matthewwardrop: The current code looks like:
```
have_pandas_categorical = (have_pandas and hasattr(pandas, "Categorical"))
if not have_pandas:
    _pandas_is_categorical_dtype = None
else:
    if hasattr(pandas, "CategoricalDtype"):  # pandas >= 0.25
        _pandas_is_categorical_dtype = lambda x: isinstance(getattr(x, "dtype", x), pandas.CategoricalDtype)
    elif hasattr(pandas, "api"):  # pandas >= 0.19
        _pandas_is_categorical_dtype = getattr(pandas.api.types, "is_categorical_dtype", None)
    else:  # pandas <=0.18
        _pandas_is_categorical_dtype = getattr(pandas.core.common,
                                               "is_categorical_dtype", None)
have_pandas_categorical_dtype = _pandas_is_categorical_dtype is not None
```

I don't see how we could ever generate a deprecation warning... so I'll just drop the warnings filter for now. It's arcane enough as it is :).

PR #213, opened 2024-11-08
explicitely support Py3.6 -> Py3.13

a-detiste:
@bashtage


matthewwardrop:
Thanks @a-detiste . Looks good to me.

PR #216, opened 2024-11-12
BUG: Remove pytest import from top of module

### LINKED ISSUE #215: pytest dependency issue
Hello,

I just saw CI pipeline of my project failed due to new 1.0.0 release. Looks like in #214 pytest import was added in [`patsy/eval.py`](https://github.com/pydata/patsy/blob/a514e0c259b062694a5645071634dd3cee337ceb/patsy/eval.py#L21). Since pytest is an optional requirement for tests, it results with `ModuleNotFoundError: No module named 'pytest'` in case patsy was installed without test dependencies.

bashtage:
Move import to within the test function and change skip method

closes #215

### DIFF
--- a/patsy/eval.py
+++ b/patsy/eval.py
@@ -18,7 +18,6 @@ import inspect
 import tokenize
 import ast
 import numbers
-import pytest
 from patsy import PatsyError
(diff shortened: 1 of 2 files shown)


matthewwardrop:
Thanks @bashtage .

PR #218, opened 2024-11-16
remove backport of collections.OrderedDict for Py2.7 support

a-detiste:
collections.OrderedDict is available since Python3.1

https://docs.python.org/3/library/collections.html#collections.OrderedDict


bashtage:
Given the minimum supported python is 3
6, are all dict ordered now?

a-detiste:
I've heard "It is an implementation detail" and should not be relied on.

bashtage:
Looks like 3.7 is the minimum python where all dicts are ordered.

bashtage:
> I've heard \"It is an implementation detail\" and should not be relied on.

It is required from 3.7. 

https://mail.python.org/pipermail/python-dev/2017-December/151283.html

a-detiste:
So merge this in now and change it again later when 3.7 is minimum

matthewwardrop:
@a-detiste : I don't have any objections to this PR, but also... this code-base is essentially dead. If things aren't broken, my general inclination is not to fix them. We want to the keep the package working, at least until more packages have successfully transitioned to `formulaic`... but no new feature development is planned.

I do appreciate you making the effort to improve things here, and I might merge it in if we do another release... which it looks like we might need to do for pandas 3 compatibility.

matthewwardrop:
@a-detiste are the compat files broken in your experience? If not, every change introduces some amount of risk of breakage for longstanding use cases, which then would require additional maintenance... And so is likely not worth it. I know this project still has many users, but it is also a "dead" project.

This project is more than an order of magnitude slower than formulaic for most common tasks, and lacks many of the cleanups and niceties of its successor.

So, in short, my goal is to keep it "undead" 😜 until it is no longer needed... But little more.

a-detiste:
From our point of vue, https://www.statsmodels.org/stable/install.html `statsmodel` need to be adapted before we can drop Patsy

matthewwardrop:
@a-detiste : statsmodels has actually been ported already by @bashtage , but I don't think a stable version has been released with it yet.

Until then, as I mentioned, the goal is to keep things working as close to *exactly* as they are for the the foreseeable future.

bashtage:
I think I could let this in if it has a path where it uses a plain dict for python 3.7+ by assigning OrderedDict = dict when at least 3.7.

PR #229, opened 2025-07-22
MAINT: Prepare patsy for pandas3 StringDtype

bashtage:
Adds support for StringDtype
Fixes tests that are not valid with copy-on-write


patsy/missing.py
```diff
@@ -183,7 +183,7 @@ def _handle_NA_drop(self, values, is_NAs, origins):
             total_mask |= is_NA
         good_mask = ~total_mask
         # "..." to handle 1- versus 2-dim indexing
-        return [v[good_mask, ...] for v in values]
+        return [v[good_mask] if v.ndim == 1 else v[good_mask, ...] for v in values]
```
matthewwardrop, line 186: Is this due to upstream indexing changes? Kind of annoying if `...` no longer supports "zero" expansion.

patsy/util.py
```diff
@@ -799,7 +820,7 @@ def test_safe_is_pandas_categorical():
 #   https://github.com/pydata/pandas/issues/9581
 #   https://github.com/pydata/pandas/issues/9581#issuecomment-77099564
 def safe_issubdtype(dt1, dt2):
-    if safe_is_pandas_categorical_dtype(dt1):
+    if safe_is_pandas_categorical_dtype(dt1) or safe_is_pandas_string_dtype(dt1):
```
matthewwardrop, line 823: Hmmm... are there other places that string dtypes should be treated as categorical, no? I'll need to take a look, since I haven't looked at patsy code for a while.

patsy/missing.py
bashtage: Yes, it appears that it is stricter and no longer supports expansion.  I only found this by running against the statsmodels test suite.

matthewwardrop: Hmmm... this still works in my testing. Is there an upstream numpy change I should be worried about? (These are, or should be, numpy arrays still).

patsy/util.py
```diff
@@ -65,6 +70,11 @@
         )
 have_pandas_categorical_dtype = _pandas_is_categorical_dtype is not None
 
+
+def _pandas_is_string_dtype(x):
```
matthewwardrop, line 74: Why do we need this private function? Can't we just consolidate this into `safe_is_pandas_string_dtype`?

patsy/util.py
matthewwardrop: Doesn't look like it, but I didn't do an exhaustive search.

matthewwardrop (commented):
Left a few questions/feedback. Once we've addressed those, let's get this in and released.

PR #260, opened 2026-05-31
fix: VarLookupDict __iter__ + DataFrame dtype compat (fixes #259, #232)

### LINKED ISSUE #232: `build._eval_factor` error condition attempts to access attribute `dtype` on `DataFrame` object, but no such attribute exists
In `_eval_factor` we have:
```python
    # Returns either a 2d ndarray, or a DataFrame, plus is_NA mask
    if factor_info.type == "numerical":
        result = atleast_2d_column_default(result, preserve_pandas=True)
```
followed by
```python
        if not safe_issubdtype(np.asarray(result).dtype, np.number):
            raise PatsyError(
                "when evaluating numeric factor %s, "
                "I got non-numeric data of type '%s'" % (factor.name(), result.dtype),
                factor,
            )
```

When `result` is a `pandas.DataFrame` but the factor is not actually a safe numeric type, this code will try to raise an error, but in doing so it attempts to access `result.dtype` which does not exist.

This is not a critical bug; it just results in some pretty confusing error messages.

### LINKED ISSUE #259: VarLookupDict should implement __iter__, causes CPython crash when using statsmodels in marimo notebook
Using statsmodels with patsy formulas in a marimo notebook causes a CPython crash (!) if a variable in the patsy formula is undefined. There is an underlying CPython bug (it shouldn't crash!), which I have described at https://github.com/python/cpython/issues/129605#issuecomment-4583354945. This is easily fixable on the patsy side, though. (Happy to submit a small PR if this sounds good to you.)

Here's the real-world reproducer (and how I came across this bug). Create a marimo (0.23.8) notebook with the following cell (that imports statsmodel 0.14.6, which in turn imports patsy 1.0.2).
```python
import polars as pl
import statsmodels.formula.api as smf

df = pl.DataFrame({"a2": range(10), "b": range(10, 20)})
model = smf.ols(
    "a ~ b",
    data=df.to_pandas(),
)
```
The CPython process running the marimo kernel will crash when the cell is run. A misspelled column name in a formula string triggers a `NameError` inside patsy's `eval()` call (where `f_locals` is a live `VarLookupDict`). Marimo calls `traceback.print_exception` directly from Python to format cell errors — unlike the REPL, which uses C-level `PyErr_Display` and is robust to this — so the `KeyError(0)` propagates uncaught and kills the notebook kernel.

Here's the minimal reproducer describing the easy fix on the patsy side:
```python
from patsy.eval import VarLookupDict

d = VarLookupDict([{"x": 1, "y": 2}])
list(d)
# Expected: ["x", "y"]
# Actual: KeyError: 0
```

kouyouqi123:
## Fixes #259 — VarLookupDict missing `__iter__`
Added `__iter__`, `__len__`, and `keys()` to `VarLookupDict` to prevent CPython crash during traceback formatting.

## Fixes #232 — DataFrame `.dtype` access in `_eval_factor`
Changed `result.dtype` to `np.asarray(result).dtype` so the error message works with both numpy arrays and pandas DataFrames.

### DIFF
--- a/patsy/build.py
+++ b/patsy/build.py
@@ -80,7 +80,8 @@ def _eval_factor(factor_info, data, NA_action):
         if not safe_issubdtype(np.asarray(result).dtype, np.number):
             raise PatsyError(
                 "when evaluating numeric factor %s, "
-                "I got non-numeric data of type '%s'" % (factor.name(), np.asarray(result).dtype),
+                "I got non-numeric data of type '%s'"
+                % (factor.name(), np.asarray(result).dtype),
                 factor,
             )
         return result, NA_action.is_numerical_NA(result)
--- a/patsy/build.py
+++ b/patsy/build.py
@@ -80,7 +80,7 @@ def _eval_factor(factor_info, data, NA_action):
         if not safe_issubdtype(np.asarray(result).dtype, np.number):
             raise PatsyError(
                 "when evaluating numeric factor %s, "
-                "I got non-numeric data of type '%s'" % (factor.name(), result.dtype),
+                "I got non-numeric data of type '%s'" % (factor.name(), np.asarray(result).dtype),
                 factor,
             )
         return result, NA_action.is_numerical_NA(result)
--- a/patsy/eval.py
+++ b/patsy/eval.py
@@ -72,6 +72,20 @@ class VarLookupDict(object):
         except KeyError:
             return default
 
+    def __iter__(self):
+        seen = set()
+        for d in self._dicts:
+            for key in d:
+                if key not in seen:
+                    seen.add(key)
+                    yield key
+
+    def __len__(self):
+        return len(list(iter(self)))
+
+    def keys(self):
+        return list(iter(self))
+
     def __repr__(self):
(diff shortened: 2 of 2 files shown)


bashtage:
This is the type of change that is considered hard in patsy.  Could you explain more as to why this is experienced when it does not come up in standard CPython?

bashtage:
Also, does this happen in standard CPython, only not get tested in patsy?

bashtage:
I think I understand this issues now.  Shouldn't the fix either go into the notebook environment or into CPython directly? These appear to be fully upstream of patsy.  

The hesitation is around the general treatment of patsy in being in maintenance mode.   This has been interpreted to mean fixes for change in CPython where essential for it to run.  

This said, statsmodels need to get out the version where formulaic is available.

bashtage:
For this to have a realistic prospect (given the status of the bug and the maintenance mode of patsy), we will need 100% coverage.  Ideally you could add a single test that triggers the error using the same API that caused this issue in the first place, and we will see that everything that has been added will be hit.

PR #267, opened 2026-07-25
Remove intercept for a bare "+ -1" / "+ 0", matching "- 1" (fixes #158)

### LINKED ISSUE #158: "+ -1" equivalence with "-1" in patsy formula?
As per patsy's [documentation](https://patsy.readthedocs.io/en/latest/formulas.html#intercept-handling), the following results in a design matrix with no intercept:

```python
>>> patsy.dmatrix('a + b - 1',
       pandas.DataFrame({'a':[1,2], 'b':[3,4]}))
```
```
DesignMatrix with shape (2, 2)
  a  b
  1  3
  2  4
  Terms:
    'a' (column 0), 'b' (column 1)
```
The documentation seems to imply that appending `+ -1` to the formula `a + b` should have the same effect as appending '-1'; however, it doesn't seem that the intercept is removed for the former:
```python
>>> patsy.dmatrix('a + b + -1',
       pandas.DataFrame({'a':[1,2],'b':[3,4]}))
```
```
DesignMatrix with shape (2, 3)
  Intercept  a  b
          1  1  3
          1  2  4
  Terms:
    'Intercept' (column 0)
    'a' (column 1)
    'b' (column 2)
```
Is the above expected?

I'm using patsy 0.5.1 with python 3.7.6 and pandas 1.0.5.

deepakganesh78:
This fixes #158.

## Root cause
`_eval_binary_plus` only propagated an **added** intercept from its right operand (via the `intercept` flag); it ignored the right operand's `intercept_removed` flag entirely, keeping only the *left* operand's intercept state.

So `a + b - 1` removes the intercept (the removal lives on the left operand), but `a + b + -1` did not — even though the [docs](https://patsy.readthedocs.io/en/latest/formulas.html#intercept-handling) imply that appending `+ -1` is equivalent to appending `- 1`. The maintainer confirmed in the issue that this is a bug and that the successor, Formulaic, treats `a + b + -1` as `a + b`.

## Fix
When the right operand of a binary `+` is a **bare** intercept literal (a chain of unary `+`/`-` ending directly in `0`/`1`, as in `-1`), honor its intercept removal — mirroring the existing direct `+ 0` (ZERO) handling just above it.

A small `_is_bare_intercept()` helper inspects the parse node so that removal is **not** propagated out of a compound/parenthesized expression such as `1 + (a + 0)`, which keeps the intercept **by design** (covered by the existing `~ 1 + (a + 0)` test, which still passes).

### Behavior after fix
```
a + b - 1      -> ['a', 'b']
a + b + -1     -> ['a', 'b']   # was ['Intercept', 'a', 'b']
a + b + 0      -> ['a', 'b']   # was ['Intercept', 'a', 'b']
a + -1 + 1     -> ['Intercept', 'a']   # re-adding still works
0 + a          -> ['a']        # unchanged
1 + (a + 0)    -> ['Intercept', 'a']   # unchanged, by design
```

## Tests
Added regression cases to `_eval_tests` and a changelog entry. Full suite is green:

```
147 passed, 1 skipped
```
(scipy installed so the spline tests run.)

### DIFF
--- a/patsy/desc.py
+++ b/patsy/desc.py
@@ -266,6 +266,19 @@ def _eval_binary_plus(evaluator, tree):
                 False,
                 left_expr.terms + right_expr.terms,
             )
+        elif right_expr.intercept_removed and _is_bare_intercept(tree.args[1]):
+            # e.g. "a + b + -1": a bare intercept literal on the right (unary
+            # minus applied directly to 1, not wrapped in parentheses) removes
+            # the intercept, consistently with "- 1".  This mirrors the direct
+            # "+ 0" (ZERO) case handled above.  We deliberately do *not*
+            # propagate intercept removal out of a compound expression such as
+            # "1 + (a + 0)", which keeps the intercept by design.
+            return IntermediateExpr(
+                False,
+                None,
+                True,
+                left_expr.terms + right_expr.terms,
+            )
         else:
             return IntermediateExpr(
                 left_expr.intercept,
@@ -273,6 +286,15 @@ def _eval_binary_plus(evaluator, tree):
                 left_expr.intercept_removed,
                 left_expr.terms + right_expr.terms,
             )
+
+
+# True if `node` is a *bare* intercept literal directly on the right of a "+",
+# i.e. a chain of unary +/- ending in a numeric literal (as in "-1"), rather
+# than a compound/parenthesized expression (as in "(a + 0)").
+def _is_bare_intercept(node):
+    while node.type in ("+", "-") and len(node.args) == 1:
+        node = node.args[0]
+    return node.type in ("ONE", "ZERO", "NUMBER")
 
 
 def _eval_binary_minus(evaluator, tree):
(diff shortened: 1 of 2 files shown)


patsy/desc.py
```diff
@@ -266,6 +266,19 @@ def _eval_binary_plus(evaluator, tree):
...
                 left_expr.terms + right_expr.terms,
             )
+        elif right_expr.intercept_removed and _is_bare_intercept(tree.args[1]):
+            # e.g. "a + b + -1": a bare intercept literal on the right (unary
+            # minus applied directly to 1, not wrapped in parentheses) removes
+            # the intercept, consistently with "- 1".  This mirrors the direct
+            # "+ 0" (ZERO) case handled above.  We deliberately do *not*
+            # propagate intercept removal out of a compound expression such as
+            # "1 + (a + 0)", which keeps the intercept by design.
+            return IntermediateExpr(
```
bashtage, line 276: In my opinion, this is a "won't fix" type of bug given where the project is.  I think a warning would be OK, though.  To fix this, one would need to go through a reasonably long deprecation cycle since this is a change to existing code, even if the original implementation is not great.  Something that explains to the user what happens here and then tells the user how they can fix it (by slightly altering their formula) is better than a silent behaviour change. 

statsmodels will be out with formulaic support next month, so this is moving to lower importance.

bashtage (changes requested):
It would be appropriate to warn and let the user know since they can easily fix it.

